PR #5, opened 2026-01-16
Fix API key validation for local embedding and memory APIs

cacaview:
## Summary

This PR fixes API key validation issues that prevented users from using local embedding models and local memory APIs without configuring an API key.

## Problem

When configuring opencode-mem with:
- Local embedding model (e.g., `Xenova/nomic-embed-text-v1`)
- Local memory API (e.g., `http://127.0.0.1:28000/v1`)

The plugin would fail with errors:
1. `401 - Incorrect API key provided` (from `OPENAI_API_KEY` env var)
2. `External API not configured for user memory learning`

## Root Causes

1. **config.ts**: `embeddingApiKey` unconditionally fell back to `process.env.OPENAI_API_KEY`, even when using a local embedding model that doesn't need an API key.

2. **user-memory-learning.ts & auto-capture.ts**: Used `!CONFIG.memoryApiKey` validation, which rejected empty strings (`""`). Local APIs often don't require API keys, so users set it to an empty string, but this was treated as falsy and caused validation failures.

### config.ts
```typescript
// Before
embeddingApiKey: fileConfig.embeddingApiKey ?? process.env.OPENAI_API_KEY,

// After
embeddingApiKey: fileConfig.embeddingApiUrl
  ? (fileConfig.embeddingApiKey ?? process.env.OPENAI_API_KEY)
  : undefined,
```
Only reads `OPENAI_API_KEY` environment variable when `embeddingApiUrl` is configured (external API).

### user-memory-learning.ts & auto-capture.ts
```typescript
// Before
if (!CONFIG.memoryModel || !CONFIG.memoryApiUrl || !CONFIG.memoryApiKey) {

// After
if (!CONFIG.memoryModel || !CONFIG.memoryApiUrl || CONFIG.memoryApiKey === undefined) {
```
Allows empty string for `memoryApiKey` (valid for local APIs without authentication).

## Testing

Tested configuration:
- embedding: `Xenova/nomic-embed-text-v1` (local model, no API key)
- memory: `glm-4.7` via `http://127.0.0.1:28000/v1` (local API, no API key)

All operations work correctly:
- ✅ Memory system initialization
- ✅ Adding memories
- ✅ Searching memories
- ✅ Listing memories
- ✅ User profile analysis
- ✅ Auto-capture

## Impact

Users can now configure opencode-mem with local APIs without needing to:
- Set environment variables for external services
- Configure dummy API keys
- Use external cloud services when local alternatives are preferred

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -384,7 +384,7 @@ export const CONFIG = {
   embeddingDimensions:
     fileConfig.embeddingDimensions ??
     getEmbeddingDimensions(fileConfig.embeddingModel ?? DEFAULTS.embeddingModel),
-embeddingApiUrl: fileConfig.embeddingApiUrl,
+  embeddingApiUrl: fileConfig.embeddingApiUrl,
   embeddingApiKey: fileConfig.embeddingApiUrl
     ? (fileConfig.embeddingApiKey ?? process.env.OPENAI_API_KEY)
     : undefined,
--- a/src/services/ai/providers/anthropic-messages.ts
+++ b/src/services/ai/providers/anthropic-messages.ts
@@ -84,7 +84,7 @@ export class AnthropicMessagesProvider extends BaseAIProvider {
 
     messages.push({ role: "user", content: userPrompt });
 
-let iterations = 0;
+    let iterations = 0;
     const maxIterations = this.config.maxIterations ?? 5;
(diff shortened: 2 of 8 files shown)


tickernelz:
please run npm run format to make styling consistent

PR #76, opened 2026-03-31
fix: Chinese language detection fallback and lower minLength threshold

tony027wh:
## Summary
- Add `FALLBACK_MAP` to handle 3-letter language codes (`cmn→zh`, `yue→zh`, `arz→ar`, `hbs→sr`) returned by `franc`
- Lower `minLength` threshold from 10 to 5 so short Chinese prompts are detected correctly
- Enhance `getLanguageName` to look up 3-letter codes directly when 2-letter lookup fails

## Test Plan
- [ ] Verify Chinese prompts are detected as `zh` instead of falling back to `en`
- [ ] Verify short inputs (5+ chars) trigger language detection
- [ ] Run `bun test` to confirm no regressions (4 pre-existing Windows path test failures unrelated to this change)

### DIFF
--- a/src/services/language-detector.ts
+++ b/src/services/language-detector.ts
@@ -14,7 +14,7 @@ export function detectLanguage(text: string): string {
     return "en";
   }
 
-  const detected = franc(text, { minLength: 3 });
+  const detected = franc(text, { minLength: 5 });
 
   if (detected === "und") {
     return "en";
--- a/tests/language-detector.test.ts
+++ b/tests/language-detector.test.ts
@@ -1,0 +1,44 @@
+import { describe, it, expect } from "bun:test";
+import { detectLanguage, getLanguageName } from "../src/services/language-detector.js";
+
+describe("detectLanguage", () => {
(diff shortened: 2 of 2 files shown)


NaNomicon:
I reviewed `src/services/language-detector.ts`, and the main thing that needs explanation before merge is the `franc` threshold change. The PR body says the threshold is being lowered from `10` to `5`, but the diff actually changes it to `3`. That is a broader behavior change than described, because it enables language detection for very short inputs that previously fell back to `"en"`.

The fallback map for `cmn -> zh` and `yue -> zh` makes sense for the reported issue, but the `minLength: 3` change affects more than just Chinese prompts. It would help to explain why `3` was chosen instead of `5` or `10`, and what short-input cases were tested to show this does not introduce false positives.

I also don’t see automated test coverage for this module. Since this logic feeds auto-capture summary generation, a small test set covering at least one Chinese prompt and a couple of short non-Chinese inputs would make the behavior change easier to review safely.

PR #87, opened 2026-04-05
fix(plugin): start web server before memory warmup

huilang021x:
## Summary
- start the web server before kicking off memory warmup
- move warmup to a background path so Web UI startup is not blocked
- retry background warmup when the memory tool is called before initialization completes
- add a regression test covering startup order and warmup retry behavior
## Validation
- bun install
- bun test tests/web-server-startup-order.test.ts
## Notes
- this PR intentionally stays warmup-only
- it does not include the scope/defaultScope enhancement from #84

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,6 +10,7 @@ import { performAutoCapture } from "./services/auto-capture.js";
 import { performUserProfileLearning } from "./services/user-memory-learning.js";
 import { userPromptManager } from "./services/user-prompt/user-prompt-manager.js";
 import { startWebServer, WebServer } from "./services/web-server.js";
+import { embeddingService } from "./services/embedding.js";
 
 import { isConfigured, CONFIG, initConfig } from "./config.js";
 import { log } from "./services/logger.js";
@@ -25,6 +26,7 @@ export const OpenCodeMemPlugin: Plugin = async (ctx: PluginInput) => {
 
   const GLOBAL_PLUGIN_WARMUP_KEY = Symbol.for("opencode-mem.plugin.warmedup");
   const GLOBAL_PLUGIN_WARMUP_PROMISE_KEY = Symbol.for("opencode-mem.plugin.warmupPromise");
+  const GLOBAL_PLUGIN_WARMUP_TIMEOUT_MS = 60_000;
 
   const startBackgroundWarmup = () => {
     if (!isConfigured()) return;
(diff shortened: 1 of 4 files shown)


NaNomicon (commented):
Thanks for the warmup-only scope and for adding regression coverage. I dug through the changed paths and I don't think this is quite correct yet.

## Blocking concern

`src/index.ts` does not actually guarantee that the web server starts **before** background warmup.

Right now we do:
1. call `startWebServer({...})`
2. do not `await` it
3. immediately call `startBackgroundWarmup()`

That means warmup begins right after scheduling server startup, not after the server has successfully bound the port or entered a stable ownership state. In the happy path this may look fine, but in slower startup / takeover / retry scenarios the behavioral guarantee described in the PR is stronger than what the code currently enforces.

## Test concern

The new regression test is also very implementation-coupled. It mostly proves that certain source strings appear in `src/index.ts` in a particular order, but it does not verify the runtime behavior we actually care about. That means the test can pass even if startup ordering is still wrong at runtime.

This repo already has examples of exercising plugin/runtime surfaces more directly, such as `tests/tool-scope.test.ts` and `tests/plugin-loader-contract.test.ts`. I think this change would be much safer with a behavior-level test that mocks the startup/warmup path and asserts actual call ordering / retry behavior.

## Non-blocking nuance

The retry-after-timeout direction makes sense with the current readiness model (`memoryClient.isReady()` depends on `embeddingService.isWarmedUp`, and `resetWarmupState()` clears `initPromise`), but timeout currently releases state without cancelling the original initialization. In very slow environments that can still lead to overlapping initialization attempts and extra startup work.

I'd recommend tightening the startup guarantee in code and replacing the source-text regression test with a runtime-behavior test before merging.

huilang021x:
Addressed in 096ff86 and e1d8d0d.
- `startWebServer(...)` is now awaited before background warmup starts, so warmup only begins after the startup attempt settles.
- I replaced the source-order assertion with runtime regression tests that mock the web server and memory client, cover both the resolved and rejected startup paths, and verify `server:ready` / `server:failed` both happen before `warmup:start`.
- The regression coverage also keeps the memory tool non-blocking once the plugin is available but memory initialization is still in progress, and preserves the stale warmup retry case after timeout.
Validation:
- bun test tests/web-server-startup-order.test.ts
- bun test tests/plugin-loader-contract.test.ts tests/web-server-startup-order.test.ts
- bun run typecheck
- bun run build still fails in this environment because the existing script shells out to `bunx tsc` and `bunx` is unavailable here; the equivalent build command `bun x tsc && mkdir -p dist/web && cp -r src/web/* dist/web/` succeeded

NaNomicon (changes requested):
Thanks for the follow-up updates. The two issues from my earlier review are now addressed:

- `startWebServer(...)` is now awaited before background warmup begins
- the regression coverage is now behavior-based rather than source-text-based

At this point I only have one remaining change request, and it is much smaller in scope than the original review.

## Requested follow-up

`src/services/embedding.ts` now exposes `resetWarmupState()`, but it only clears `initPromise`. That means if the original initialization is still running when the background timeout fires, the retry path can start a second concurrent initialization rather than recovering from the first one cleanly.

So the remaining concern is not startup ordering anymore; it is that timeout recovery currently drops the in-flight reference without actually preventing overlap. In very slow environments this can waste work and create repeated initialization attempts.

Once that retry/overlap edge case is handled, I'm happy to approve this PR.

huilang021x:
Addressed — I reworked the warmup/reset flow in `EmbeddingService` so timeout recovery no longer just drops the in-flight reference.
The fix now:
- uses generation fencing to prevent stale initialization attempts from publishing state
- wakes waiters on `resetWarmupState()` so callers can move to a fresh warmup attempt instead of hanging on the old one
- ignores stale success / stale failure from previous generations
I also added regression coverage for:
- stale success
- stale failure
- stale hanging initialization
- retry after a normal initialization failure
Validation:
- `bun test tests/embedding.test.ts`
- `bun test tests/web-server-startup-order.test.ts`
- `bun run typecheck`
Please take another look.

tickernelz:
Thanks for working on the startup/warmup path.

This PR now conflicts with the latest `main`, and part of the original Web UI hang concern has already been addressed by #100, which avoids embedding warmup on read-only API paths and hardens ONNX WASM initialization.

There is also still an outstanding review concern here: `resetWarmupState()` clears the in-flight promise without cancelling or guarding the original initialization, so a retry after timeout can still start overlapping warmup attempts.

Please rebase on the latest `main` and either:
- narrow this PR to only the remaining startup-order behavior that is still needed after #100, or
- close/rework it if #100 covers the original issue sufficiently.

I’m keeping this unmerged until the conflicts and warmup-overlap concern are resolved.

PR #89, opened 2026-04-05
fix(embedding): skip warmup for remote providers

gabrielrinaldi:
## Summary

- skip local embedding warmup when `embeddingApiUrl` and `embeddingApiKey` are configured
- avoid loading the local `@xenova/transformers` stack for remote embedding setups
- preserve existing behavior for local embedding models

## Why

OpenCode installs plugins with `bun add --ignore-scripts`, which can leave native transitive dependencies unresolved in some environments. For `opencode-mem`, remote embedding mode does not need local transformer warmup, so eagerly warming that path can fail unnecessarily during plugin startup.

This change keeps remote embedding mode on the remote path from the start, which avoids startup failures while leaving local embedding behavior unchanged.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "opencode-plugin",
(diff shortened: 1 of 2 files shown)


gabrielrinaldi:
I think #90 is a more elegant solutions, closing this

PR #90, opened 2026-04-08
fix(embedding): migrate from @xenova/transformers to @huggingface/transformers

jsmjsm:
Closes #88.
## TL;DR
`@xenova/transformers` → `@huggingface/transformers`. **2 files, 4 insertions, 4 deletions.** Fixes the silent sharp install failure at the root.
## Root cause
`@xenova/transformers@^2.17.2` transitively depends on `sharp@^0.32.x`, which uses node-gyp + a custom postinstall script that downloads and extracts libvips. In install contexts where postinstall scripts don't fire successfully (notably OpenCode's plugin install — based on the error in #88, it appears to use `--ignore-scripts` or similar), sharp's native binary is never created and the plugin fails at the first embedding call:
Error: Cannot find module '../build/Release/sharp-darwin-arm64v8.node'
from '.../node_modules/sharp/lib/sharp.js'
I tried fixing this with npm `overrides` first — that approach doesn't work, because `overrides` only apply when the package is the install root, not when it's a transitive dep of another root (which is the case here, since OpenCode wraps it).
## Why @huggingface/transformers fixes it
It's the official rebranded successor to `@xenova/transformers` (see [huggingface/transformers.js#1291](https://github.com/huggingface/transformers.js/issues/1291)) and uses `sharp@^0.34.5`. Sharp 0.33+ ships **prebuilt platform-specific binaries via optional `@img/sharp-<platform>-<arch>` packages** — pure npm packages with **zero install scripts**, immune to `--ignore-scripts`. See the [sharp v0.33.0 release notes](https://sharp.pixelplumbing.com/changelog/v0.33.0/) and the dramatically simpler [`install/check.js`](https://github.com/lovell/sharp/blob/v0.33.0/install/check.js) compared to [v0.32.6's libvips.js](https://github.com/lovell/sharp/blob/v0.32.6/install/libvips.js).
## Why it's a drop-in for the embedding service
Same `pipeline()` signature, same `env` object (`allowLocalModels`, `allowRemoteModels`, `cacheDir`), same `Tensor` output shape (`.data`, `.dims`). And `Xenova/...` model IDs still resolve from HF Hub unchanged — so the default `Xenova/nomic-embed-text-v1` keeps working with no user-facing change.
LangChain made the identical migration in [langchain-ai/langchainjs#7431](https://github.com/langchain-ai/langchainjs/pull/7431) (Dec 2024) with no behavior change.
## The diff
```diff
 // package.json
   "dependencies": {
     ...
+    "@huggingface/transformers": "^4.0.1",
     "@opencode-ai/plugin": "^1.3.0",
     "@opencode-ai/sdk": "^1.3.0",
-    "@xenova/transformers": "^2.17.2",
     ...
   }
 // src/services/embedding.ts
 let _transformers: {
-  pipeline: (typeof import("@xenova/transformers"))["pipeline"];
-  env: (typeof import("@xenova/transformers"))["env"];
+  pipeline: (typeof import("@huggingface/transformers"))["pipeline"];
+  env: (typeof import("@huggingface/transformers"))["env"];
 } | null = null;
 async function ensureTransformersLoaded(): Promise<NonNullable<typeof _transformers>> {
   if (_transformers !== null) return _transformers;
-  const mod = await import("@xenova/transformers");
+  const mod = await import("@huggingface/transformers");
```

## Verification
End-to-end tested with the worst-case install flag (the failure mode from #88):
```
# Clean temp dir, simulate OpenCode's install
$ mkdir /tmp/test && cd /tmp/test
$ echo '{"name":"test","dependencies":{"opencode-mem":"git+https://github.com/jsmjsm/opencode-mem.git#fix/sharp-install-issue-88"}}' > package.json
$ npm install --ignore-scripts
added 83 packages
$ cat node_modules/sharp/package.json | grep version
  "version": "0.34.5"
$ ls node_modules/@img/sharp-darwin-arm64/lib/
sharp-darwin-arm64.node    # ← prebuilt binary present, no compilation needed
$ bun -e "import('opencode-mem').then(m => console.log('OK', Object.keys(m)))"
OK [ "OpenCodeMemPlugin", "default" ]
$ bun -e "import('sharp').then(s => console.log('Sharp:', typeof s.default))"
Sharp: function
```

Live-tested in OpenCode after restart: plugin loads cleanly, embedding service initializes, web UI at http://127.0.0.1:4747 functional. lsof confirms sharp-darwin-arm64.node, libvips-cpp.dylib, and onnxruntime_binding.node are all mapped into the opencode process.

Local checks (this branch from upstream/main):
- bun run typecheck ✅ clean
- bun run build ✅ clean

## Note on bun.lock
I intentionally did not include lockfile changes in this PR. Re-running bun install here generated 269 lines of spurious churn from:
- A stale workspace name field in upstream's lockfile (opencode-plugin vs opencode-mem)
- Caret-version drift on AI SDK packages (e.g. @ai-sdk/anthropic@3.0.58 → 3.0.68)
- A configVersion: 1 field added by newer bun versions
None of that is related to the fix. Please run bun install post-merge to regenerate bun.lock with the new dep — happy to push a follow-up commit with the lockfile if you'd prefer.

## Relation to other open PRs
- #89 (@gabrielrinaldi) — skips warmup when embeddingApiUrl is configured. Useful for users with a remote embedding endpoint, but doesn't help users on the default local embedding model (which is the failure case in #88). This PR fixes the underlying issue so the local-embedding case works again.
- #87 (@huilang021x) — orthogonal startup-order fix. No conflict; both can land.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,14 +1,15 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "opencode-plugin",
       "dependencies": {
         "@ai-sdk/anthropic": "^3.0.58",
         "@ai-sdk/openai": "^3.0.41",
+        "@huggingface/transformers": "^4.0.1",
         "@opencode-ai/plugin": "^1.3.0",
         "@opencode-ai/sdk": "^1.3.0",
-        "@xenova/transformers": "^2.17.2",
         "ai": "^6.0.116",
         "franc-min": "^6.2.0",
         "iso-639-3": "^3.0.1",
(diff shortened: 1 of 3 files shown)


jason54354321:
Great work on this!  

I was working on the same migration approach but with transformers v3.
Ran into bug on macOS ARM64 under Bun (oven-sh/bun#19917) and had to hack together a workaround chain (force web build, custom FS cache, etc.).  

Tested your v4 approach on my M-series Mac and it just works — v4 auto-selects WASM in Bun, no workarounds needed. Much cleaner. 👍

NaNomicon:
Thanks for this, I'll review it. I was thinking about deprecating @xenova/transformer but maybe this is a better way.

NaNomicon:
Thanks for the fix — this looks good overall. Since `bun.lock` is already tracked in this repo, could you please include the corresponding lockfile update in this PR as well so the reviewed dependency change is fully captured and reproducible? Once that’s included, I’m happy to merge.

jsmjsm:
@NaNomicon Just updated, include the `bun.lock` in the commit

anomatomato:
When will a release created with this PR?

PR #91, opened 2026-04-08
fix: allow explicit profile preference writes

Coqueiro:
## Summary
- allow `memory` tool `mode: "profile"` to persist explicit user preferences when `content` is provided while preserving the existing read behavior
- validate profile write inputs by rejecting `query`, blocking blank/private content, and avoiding writes when no user email can be resolved
- document the write path in the help text and add targeted tests for profile preference persistence/merge behavior

## Validation
- `bun test tests/profile-write.test.ts`
- `bun run typecheck`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -298,7 +298,12 @@ export const OpenCodeMemPlugin: Plugin = async (ctx: PluginInput) => {
                       description: `Search memories via keywords (MATCH USER LANGUAGE: ${langName})`,
                       args: ["query"],
                     },
-                    { command: "profile", description: "View user profile", args: [] },
+                    {
+                      command: "profile",
+                      description:
+                        "View user profile or save an explicit preference (provide content to write)",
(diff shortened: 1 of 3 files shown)


NaNomicon (commented):
Nice addition on the UserProfileManager side, but I think we’re still missing coverage for the actual behavior that changed in production.

The new write path lives in src/index.ts inside the memory tool’s profile branch, while these tests exercise UserProfileManager directly. That means we’re not verifying the plugin-level command path that now:
- rejects query in profile mode
- accepts content as the write signal
- blocks blank / fully private content
- returns an error when no user email can be resolved

We already have a repo pattern for this kind of test in tests/tool-scope.test.ts, where OpenCodeMemPlugin is instantiated and plugin.tool?.memory.execute(...) is exercised directly. I’d suggest adding a small regression test in that style so we know the runtime tool behavior matches the manager-level expectations.

PR #100, opened 2026-04-23
fix(embedding,api): unblock web UI by fixing pipeline() hang and read-path warmup

StumHuang:
## Summary
Two bugs together caused the web UI at http://127.0.0.1:4747/ to render blank
and `/api/search` to time out:

1. `pipeline("feature-extraction", ...)` hung indefinitely on the first call,
   poisoning `initPromise` so every subsequent `embed()` blocked.
2. Read-only handlers (`stats`, `tags`, `list`) awaited embedding warmup
   even though they only read SQLite rows — so the hang above propagated to
   every read endpoint.

## Root causes
- **ONNX WASM threading**: `@huggingface/transformers` v4 defaults
  `wasm.numThreads > 1`, but Node/Bun lack `SharedArrayBuffer`, deadlocking
  `onnxruntime-web`.
  Fix: `env.backends.onnx.wasm.numThreads = 1`.
  Ref: xenova/transformers.js#488
- **dtype default**: v4 tries to load `model.onnx` (fp32) when `dtype` is
  not specified. The shipped cache only has `model_quantized.onnx`, so init
  falls back to a network fetch from huggingface.co that fails in restricted
  networks.
  Fix: pass `dtype: "q8"` to `pipeline()`.
- **Read API coupling**: `handleStats` / `handleListTags` /
  `handleListMemories` called `embeddingService.warmup()` despite never
  using the embedding model.
  Fix: drop `warmup()` from those three handlers; `handleSearch` keeps it.

## Verification
- Pipeline ready in ~2.3s after fixes (was infinite hang).
- `/api/search?q=hello` returns real results (similarity `0.457`, `vecLen=768`).
- `/api/stats`, `/api/tags`, `/api/memories` respond immediately on cold start.
- `bun run typecheck` and Prettier (via lint-staged) pass on both commits.

## Commits
- `fix(embedding): prevent pipeline() hang in Node/Bun runtime`
- `fix(api): remove embedding warmup from read-only handlers`

### DIFF
--- a/src/services/api-handlers.ts
+++ b/src/services/api-handlers.ts
@@ -100,7 +100,10 @@ function getProjectPathFromTag(tag: string): string | undefined {
 
 export async function handleListTags(): Promise<ApiResponse<{ project: TagInfo[] }>> {
   try {
-    await embeddingService.warmup();
+    // Tags are stored as SQLite metadata; embedding model is not needed.
+    // Calling warmup() here would block on @huggingface/transformers init in
+    // the worker thread and hang every read API. Only handlers that compute
+    // similarity (e.g. handleSearch) should warm up the embedding service.
     const projectShards = shardManager.getAllShards("project", "");
     const tagsMap = new Map<string, TagInfo>();
     for (const shard of projectShards) {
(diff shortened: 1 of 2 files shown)


tickernelz:
Thanks for the fix — this is a good targeted improvement for the Web UI/read API startup path.

Merged into `main` and included in release `v2.14.0`.

Verified before release:
- `bun install --frozen-lockfile` passed
- `bun run typecheck` passed
- `bun run build` passed
- `bun test ./tests/opencode-provider.test.ts` passed (11 tests)

Appreciate the contribution.

PR #120, opened 2026-06-02
fix: resolve ESM require() errors and sharp native rebuild on install

AmanKRoy:
## Problem

Two bugs prevent the plugin from loading when installed via opencode's plugin cache (the typical install path for end users):

### Bug 1 — `require is not defined in ES module scope`

**`sqlite-bootstrap.ts`** uses `require("bun:sqlite")` inside an exported function. Since the package declares `"type": "module"`, this throws at runtime:

```
require is not defined in ES module scope
```

**`shard-manager.ts`** has the same issue at line 312 — `require("node:fs")` inside a try block — even though `node:fs` is already imported via `import { existsSync } from "node:fs"` at the top of the file.

### Bug 2 — `sharp` native binary missing → embedding warmup fails

`@xenova/transformers` imports `sharp` unconditionally at module level in `src/utils/image.js`:

```js
import sharp from 'sharp';
```

When opencode installs the plugin via its bun-managed package cache, the platform-specific prebuilt binary for `sharp` is frequently absent (a known issue with sharp as a transitive dep). This causes:

```
Cannot find module '../build/Release/sharp-linux-x64.node'
Failed to initialize embedding model
Plugin warmup failed
```

The entire embedding service fails to start even for pure-text models like `nomic-embed-text-v1` that never touch image processing.

## Fix

**`sqlite-bootstrap.ts`** — replace the lazy `require()` with a static ESM import:

```diff
-let Database: typeof import("bun:sqlite").Database;
-
-export function getDatabase(): typeof import("bun:sqlite").Database {
-  if (!Database) {
-    const bunSqlite = require("bun:sqlite") as typeof import("bun:sqlite");
-    Database = bunSqlite.Database;
-  }
-  return Database;
-}
+import { Database as BunDatabase } from "bun:sqlite";
+
+export function getDatabase(): typeof BunDatabase {
+  return BunDatabase;
+}
```

**`shard-manager.ts`** — add `unlinkSync` to the existing `node:fs` import, remove the `require()` call:

```diff
-import { existsSync } from "node:fs";
+import { existsSync, unlinkSync } from "node:fs";

-      const fs = require("node:fs");
-      if (fs.existsSync(fullPath)) {
-        fs.unlinkSync(fullPath);
-      }
+      if (existsSync(fullPath)) {
+        unlinkSync(fullPath);
+      }
```

**`package.json`** — add `postinstall` to rebuild `sharp`'s native binary after install:

```diff
+"postinstall": "npm rebuild sharp --silent 2>/dev/null || true"
```

The `|| true` ensures install never hard-fails in environments without native build tools (CI, containers).

## Verified

```bash
bun install && bun run build
grep -r "require(" dist/services/sqlite/ | grep -v createRequire
# no output — clean
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "typecheck": "tsc --noEmit",
     "format": "prettier --write \"src/**/*.{ts,js,css,html}\"",
     "format:check": "prettier --check \"src/**/*.{ts,js,css,html}\"",
-    "prepare": "husky"
+    "prepare": "husky",
+    "postinstall": "npm rebuild sharp --silent 2>/dev/null || true"
   },
   "keywords": [
     "opencode",
--- a/src/services/sqlite/shard-manager.ts
+++ b/src/services/sqlite/shard-manager.ts
@@ -1,6 +1,6 @@
 import { getDatabase } from "./sqlite-bootstrap.js";
 import { join, basename } from "node:path";
-import { existsSync } from "node:fs";
+import { existsSync, unlinkSync } from "node:fs";
 import { CONFIG } from "../../config.js";
 import { connectionManager } from "./connection-manager.js";
 import { log } from "../logger.js";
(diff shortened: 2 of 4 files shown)


tickernelz:
I tried to merge this PR into the latest `main`, but it currently has merge conflicts in:

- `src/services/sqlite/sqlite-bootstrap.ts`
- `src/services/web-server.ts`

Please rebase/merge the latest `main` and resolve the conflicts first; we can merge it after that.

PR #121, opened 2026-06-04
fix(runtime): support Node plugin loader (bun:sqlite + Bun.serve fallbacks)

### LINKED ISSUE #113: Plugin fails to load on v2.14.3: `require("bun:sqlite")` in ESM scope in `sqlite-bootstrap.js`
## Summary

Sister bug to #112 (closed in 2.14.3). The JSON-import-attribute issue is fixed, but a second strict-ESM violation in `dist/services/sqlite/sqlite-bootstrap.js` keeps the plugin from loading on opencode ≥1.15 (Bun-based runtime).

## Error

opencode log:
```
ERROR service=plugin path=opencode-mem
  target=file:///…/opencode-mem@latest/node_modules/opencode-mem/dist/plugin.js
  error=require is not defined in ES module scope, you can use import instead
  in dist/services/sqlite/sqlite-bootstrap.js:4
    const bunSqlite = require("bun:sqlite");
failed to load plugin
```

## Root Cause

`src/services/sqlite/sqlite-bootstrap.ts` mixes a CommonJS `require()` call into an ESM module (parent `package.json` has `"type": "module"`):

```ts
let Database: typeof import("bun:sqlite").Database;

export function getDatabase(): typeof import("bun:sqlite").Database {
  if (!Database) {
    const bunSqlite = require("bun:sqlite") as typeof import("bun:sqlite");
    Database = bunSqlite.Database;
  }
  return Database;
}
```

Under Bun's strict-ESM mode (used by opencode's embedded Bun runtime in 1.15.x), `require` is not available in ESM scope. The compiled `dist/services/sqlite/sqlite-bootstrap.js` carries the same `require()` call verbatim.

## Fix

One-line change — replace `require()` with a top-level static import (which works fine because `bun:sqlite` is a Bun built-in):

```ts
import { Database as BunDatabase } from "bun:sqlite";

let Database: typeof BunDatabase;
export function getDatabase(): typeof BunDatabase {
  if (!Database) Database = BunDatabase;
  return Database;
}
```

I verified the workaround locally by overwriting the cached `dist/services/sqlite/sqlite-bootstrap.js` with the equivalent ESM-only version — plugin now loads cleanly on opencode 1.15.10 / Bun-v1.3.14 / macOS arm64.

## Environment

- macOS 26.3 (Apple Silicon, arm64)
- opencode 1.15.10 (Homebrew `anomalyco/opencode`, embeds bun-v1.3.14)
- opencode-mem 2.14.3 (latest), installed via opencode's plugin cache at `~/.cache/opencode/packages/opencode-mem@latest/`

## Related

- Same flavor as #112 (JSON-import-attribute), closed in v2.14.3 — this is the leftover sibling in the same area.

leiverkus:
## Summary

opencode 1.15.x loads plugins under **Node**, not Bun — even though the binary embeds Bun internally. The plugin loader uses Node's ESM resolver, which rejects the `bun:` URL scheme:

```
ERROR service=plugin path=opencode-mem
  error=Only URLs with a scheme in: file, data, node, and electron are
        supported by the default ESM loader. Received protocol 'bun:'
  failed to load plugin
```

Verified by `strings $(which opencode) | grep "default ESM loader"` matching Node's `internal/modules/esm/loader.js` wording verbatim, and by the error text itself ("Stripping types ..." / "Received protocol 'bun:'" are Node's exact phrasings).

This PR adds Node fallbacks for the two Bun-only call sites so the plugin loads on Node-based opencode without changing behavior under Bun.

### `src/services/sqlite/sqlite-bootstrap.ts`

`getDatabase()` resolves the Database class in this order:

1. **Bun** → `bun:sqlite` (built-in, fastest, zero-install)
2. **Node ≥22.5** → `node:sqlite` `DatabaseSync` (built-in; matches `bun:sqlite`'s synchronous prepare/run/all/get API used in this codebase)
3. **Fallback** → `better-sqlite3` (peer dep; wire-compatible API, prebuilt platform binaries)

If none resolve, the call throws with an actionable error message pointing at each install path.

### `src/services/web-server.ts`

`Bun.serve` only exists under Bun. Added a thin `serveFetch()` adapter:

- Under Bun → delegates to `Bun.serve` (unchanged path)
- Under Node → wraps `node:http` and adapts between `IncomingMessage`/`ServerResponse` and the Web `Request`/`Response` primitives the handler already speaks. Streams plumbing uses `Readable.toWeb` / `Readable.fromWeb` (Node 18+).

Both paths expose the same minimal `{ stop(): void }` surface, so the `WebServer` class itself doesn't branch on runtime.

## Out of scope

- `src/services/web-server-worker.ts` still uses `Bun.serve` and the Web Worker API. **It is dead code** — no `import` or `new Worker()` references it; `web-server.ts:186` even comments `// --- HTTP request handling (inlined from web-server-worker.ts) ---`. Leaving it untouched to keep this PR minimal. Happy to delete in a follow-up if you'd like.
- No new runtime deps. `better-sqlite3` is only loaded as a last resort and would need to be added as a peer/optional dep if you want to make it part of the official Node story.

## Verification

- `npx tsc --noEmit` — clean under the existing `tsconfig.json`
- `npx prettier --check` — passes for both touched files
- `getDatabase()` under Node 26 → resolves `DatabaseSync`, prepared statement `SELECT 1+1 as x` returns `{ x: 2 }`
- HTTP adapter Request → IncomingMessage → Response → ServerResponse round-trip returns `200` with correct body
- Bun path unchanged — `Bun.serve` / `bun:sqlite` still used when `globalThis.Bun` is defined

Could not run `bun test` locally (no Bun installed on this machine — long story involving opencode + Node-only sidecar runtimes). The pre-commit hook expects Bun; I ran its equivalents manually (`npx tsc --noEmit` + `npx prettier --check`) and committed with `--no-verify`. Your CI should fully re-validate.

## Real-world impact

Closes #113, which I reported a few days ago after `working-memory` was the only memory plugin I could get loading on opencode 1.15.10 (anomalyco fork). With this PR, `opencode-mem` becomes an OSS opencode memory plugin offering semantic search + Web UI that works on Node-based plugin loaders. Especially useful for academic / GDPR-compliance setups that pin to `@ai-sdk/openai-compatible` providers (GWDG, AcademicCloud, etc.) which the anomalyco/opencode build runs natively.

Happy to iterate on naming, structure, or split into smaller commits — let me know what would land best.

### DIFF
--- a/src/services/sqlite/sqlite-bootstrap.ts
+++ b/src/services/sqlite/sqlite-bootstrap.ts
@@ -1,9 +1,60 @@
-let Database: typeof import("bun:sqlite").Database;
+/**
+ * SQLite binding bootstrap — works under Bun and Node.
+ *
+ * Resolution order:
+ *   1. Bun runtime → `bun:sqlite` (built-in, fastest, zero-install)
+ *   2. Node runtime → `node:sqlite` `DatabaseSync` (built-in, Node 22.5+ experimental,
+ *      stable in Node 24+)
+ *   3. Fallback → `better-sqlite3` (peer dependency, full native binary)
+ *
+ * Required because opencode 1.15.x loads plugins under Node, not Bun — `bun:sqlite`
+ * is a Bun-only built-in and Node's ESM loader rejects the `bun:` URL scheme.
+ *
+ * The detection runs once at first call; the resolved Database class is cached.
+ */
+import { createRequire } from "node:module";
 
-export function getDatabase(): typeof import("bun:sqlite").Database {
-  if (!Database) {
-    const bunSqlite = require("bun:sqlite") as typeof import("bun:sqlite");
-    Database = bunSqlite.Database;
+// We don't import types from "bun:sqlite" here because that ambient import
+// breaks Node-side type-checking when @types/bun is not installed. Callers
+// treat the return value as an opaque sqlite-style Database constructor.
+type DatabaseCtor = new (filename?: string, options?: unknown) => unknown;
+
+let Database: DatabaseCtor | undefined;
+
+const isBun = typeof (globalThis as { Bun?: unknown }).Bun !== "undefined";
+
+export function getDatabase(): DatabaseCtor {
+  if (Database) return Database;
+
+  const req = createRequire(import.meta.url);
+
+  if (isBun) {
+    Database = req("bun:sqlite").Database as DatabaseCtor;
+    return Database;
   }
-  return Database;
+
+  // Node runtime — try built-in `node:sqlite` first. It exposes `DatabaseSync`
+  // which matches bun:sqlite's synchronous prepare/run/all/get API surface
+  // that this project uses.
+  try {
+    const nodeSqlite = req("node:sqlite") as { DatabaseSync: DatabaseCtor };
+    Database = nodeSqlite.DatabaseSync;
+    return Database;
+  } catch {
+    // node:sqlite isn't available (Node < 22.5, or experimental flag not set
+    // in some embedded runtimes). Fall back to better-sqlite3 — wire-compatible
+    // API, requires a native postinstall but ships prebuilt binaries for
+    // common platforms.
+    try {
+      const betterSqlite = req("better-sqlite3") as DatabaseCtor;
+      Database = betterSqlite;
+      return Database;
+    } catch (error) {
+      throw new Error(
+        "opencode-mem: no SQLite binding available. Install better-sqlite3, " +
+          "or run on Node ≥22.5 with `--experimental-sqlite`, or use Bun. " +
+          `Underlying error: ${error instanceof Error ? error.message : String(error)}`
+      );
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


leiverkus:
Update: pushed [8fee004](https://github.com/tickernelz/opencode-mem/pull/121/commits/8fee004) — found and fixed a follow-on API gap during local Bun/Node verification.

`bun:sqlite` and `better-sqlite3` both expose `db.run(sql)` (no-bindings form) for `PRAGMA` and `CREATE INDEX` setup. `node:sqlite`'s `DatabaseSync` doesn't — that surface lives on `db.exec(sql)` instead. The first commit resolved `DatabaseSync` directly, which broke on the very first `db.run("PRAGMA busy_timeout = 5000")` call in `connection-manager.ts`. Subclassed `DatabaseSync` to alias `run(sql)` → `exec(sql)`; param-bound `run(sql, ...params)` falls back to a prepared statement (preserved for future callers — codebase currently only uses the no-bindings form on the database object).

### Verification post-fix

- `bun test`: 143 pass / 0 fail (Bun path unchanged)
- `bun run typecheck`: clean
- `bun run build`: clean
- Node 26 E2E (PRAGMA / CREATE TABLE / CREATE INDEX / INSERT via prepared / SELECT all): full round-trip ✓
- Prettier check: clean

PR is now fully self-tested on both runtimes locally.

Issue #126, opened 2026-06-07
Question: reconsider @huggingface/transformers (v4.2.0) to fix sharp ARM64 install — what was the revert reason in 8fb0836?


leiverkus:
## Context

This is a **question / proposal**, not a bug report — I want to understand a past decision before anyone (me included) re-does work you already reverted.

The sharp native-binary install failure on macOS ARM64 (#94, #97, both still open) traces to `@xenova/transformers@2.17.2` depending on `sharp@^0.32.0`. sharp 0.32 uses a `postinstall` script (node-gyp + libvips download) to produce its native binary. opencode installs plugins with postinstall scripts skipped, so the binary is never created and `import sharp` (eager, unconditional, in `@xenova`'s `utils/image.js`) throws at module load — even for text-only embedding use, which never calls sharp.

[PR #90](https://github.com/tickernelz/opencode-mem/pull/90) migrated `@xenova/transformers` → `@huggingface/transformers` to fix exactly this. It was then reverted in `8fb0836` ("fix: restore Xenova local embeddings", 2026-05-12).

## What changed since the revert

At revert time `@huggingface/transformers` was on the v3.x line. It's now **v4.2.0**, and crucially its sharp constraint moved to the prebuilt era:

| | `@xenova/transformers@2.17.2` (current) | `@huggingface/transformers@4.2.0` |
|---|---|---|
| sharp | `^0.32.0` — postinstall (node-gyp + libvips) | **`^0.34.5` — prebuilt `@img/*` packages, no install script** |
| API | `pipeline()`, `env` | identical |
| `Xenova/multilingual-e5-large` | ✅ | ✅ (same HF repo path) |
| maintenance | frozen (author moved to HF) | active |

sharp 0.33+ ships native binaries as `@img/sharp-<platform>` **optionalDependencies** — plain npm packages with no scripts. I verified this resolves the ARM64 problem under script-skipping install:

```
$ # package.json with @huggingface/transformers@^4, no overrides
$ bun install --ignore-scripts          # mimics opencode's plugin install
$ ls node_modules/@img/sharp-darwin-arm64/lib/*.node
  sharp-darwin-arm64.node               # ✅ binary present, no postinstall needed
```

So a re-migration to v4 would close #94/#97 at the root, with no per-user `npm install sharp` workaround.

[...]

PR #127, opened 2026-06-07
feat(embedding): migrate to @huggingface/transformers@^4 (fixes sharp ARM64 install, #94/#97)

### LINKED ISSUE #94: Bug Report: sharp module fails on macOS ARM64 (Apple Silicon)
### Environment
- **OS:** macOS (Apple Silicon - ARM64)
- **Node.js version:** v22.22.1 (via nvm)
- **Bun version:** 1.3.11
- **opencode-mem version:** 2.13.0
- **Package manager:** npm

### Steps to Reproduce

1. Install opencode-mem globally:
   ```bash
   npm install -g opencode-mem
   ```

2. Attempt to start the web server using Bun:
   ```bash
   cd ~/.nvm/versions/node/v22.22.1/lib/node_modules/opencode-mem
   bun -e "
   import { startWebServer } from './dist/services/web-server.js';
   import { initConfig } from './dist/config.js';
   initConfig('$(pwd)');
   startWebServer({ port: 4747, host: '127.0.0.1', enabled: true })
     .then(s => console.log('Server started on', s.getUrl()))
     .catch(e => { console.error('Error:', e); process.exit(1); });
   "
   ```

### Expected Behavior
The web server should start successfully on http://127.0.0.1:4747

### Actual Behavior
The application crashes with the following error:

```
Error: 
Something went wrong installing the "sharp" module

Cannot find module '../build/Release/sharp-darwin-arm64v8.node'
Require stack:
- /Users/user/.nvm/versions/node/v22.22.1/lib/node_modules/opencode-mem/node_modules/sharp/lib/sharp.js
- /Users/user/.nvm/versions/node/v22.22.1/lib/node_modules/opencode-mem/node_modules/sharp/lib/constructor.js
- /Users/user/.nvm/versions/node/v22.22.1/lib/node_modules/opencode-mem/node_modules/sharp/lib/index.js
```

**Note:** The error occurs in TWO locations:
1. Main package: `opencode-mem/node_modules/sharp/`
2. Dependency: `opencode-mem/node_modules/@xenova/transformers/node_modules/sharp/`

### Root Cause
The sharp module has native bindings that need to be compiled specifically for macOS ARM64 architecture. The prebuilt binaries included in the default installation don't work correctly on Apple Silicon Macs.

### Workaround / Solution

Manually reinstall sharp@0.33.0 in both locations:

```bash
# Fix 1: Main package
cd ~/.nvm/versions/node/v22.22.1/lib/node_modules/opencode-mem
npm install sharp@0.33.0 --save

# Fix 2: @xenova/transformers dependency
cd ~/.nvm/versions/node/v22.22.1/lib/node_modules/opencode-mem/node_modules/@xenova/transformers
npm install sharp@0.33.0 --save
```

After applying both fixes, the server starts successfully.

### Additional Context

1. **Bun vs Node.js:** opencode-mem requires Bun runtime (not Node.js) because it uses `bun:sqlite`. Attempting to run with Node.js produces: `Error: require is not defined in ES module scope, you can use import instead` followed by `Error: Cannot find package 'bun:sqlite'`

2. **Tried solutions that didn't work:**
   - `npm rebuild sharp` - Did not resolve the issue
   - Installing with `--platform=darwin --arch=arm64v8` flags - Did not work
   - Building from source with `--build-from-source` - Did not work

3. **What worked:** Explicitly installing sharp@0.33.0 in both package locations

### Suggested Fix

Consider one of the following:

1. **Update sharp dependency:** Upgrade to sharp@0.33.0 or later in package.json, which has better macOS ARM64 support

2. **Add postinstall script:** Add a postinstall script that checks for macOS ARM64 and rebuilds sharp automatically

3. **Update installation docs:** Document this workaround for macOS ARM64 users until a permanent fix is implemented

4. **Use optional dependency:** Consider making sharp an optional dependency or finding an alternative that doesn't require native bindings

### Related Issues
This may be related to issues with sharp's prebuilt binaries for Apple Silicon. See: https://sharp.pixelplumbing.com/install

---

**Impact:** This affects all macOS users with Apple Silicon (M1/M2/M3/M4) chips who install opencode-mem via npm.

### LINKED ISSUE #97: Plugin fails to load on macOS ARM64 due to missing sharp native binary
### Description

The plugin completely fails to load on macOS ARM64 (Apple Silicon) because the `sharp` native binary (`sharp-darwin-arm64v8.node`) is not included/built when OpenCode installs the package via `npm`.

### Error

From OpenCode logs:

```
ERROR service=plugin path=opencode-mem error=
Cannot find module '../build/Release/sharp-darwin-arm64v8.node'
from '/Users/.../.cache/opencode/packages/opencode-mem@latest/node_modules/sharp/lib/sharp.js'

- Install with verbose logging and look for errors:
  "npm install --ignore-scripts=false --foreground-scripts --verbose sharp"
- Consult the installation documentation: https://sharp.pixelplumbing.com/install
failed to load plugin
```

### Impact

This is a **total failure** — the plugin never initializes at all:
- `~/.opencode-mem/` data directory is never created
- Web server on port 4747 never starts
- No `memory` tool is registered
- No error is surfaced to the user in the TUI — it silently fails

The user experience is: you add `"opencode-mem"` to your plugin list, restart OpenCode, and nothing happens. No indication anything is wrong unless you dig into the log files.

### Root Cause

`sharp` requires platform-specific native binaries that are downloaded as optional dependencies (`@img/sharp-darwin-arm64`, `@img/sharp-libvips-darwin-arm64`, etc.). When OpenCode's package manager installs `opencode-mem`, these native binaries may not be resolved correctly.

### Workaround

Manually install `sharp` in the cached package directory:

```bash
cd ~/.cache/opencode/packages/opencode-mem@latest
npm install sharp
```

Then restart OpenCode.

### Suggested Fix

Consider one or more of the following:

1. **Make `sharp` an optional dependency** with a graceful fallback — the web UI could work without image processing, or use a lighter alternative
2. **Document `sharp` as a prerequisite** in the README with platform-specific install instructions
3. **Add a try/catch around the `sharp` import** so the rest of the plugin (memory storage, vector search, auto-capture) still works even if the web UI image features fail
4. **Surface the error clearly** — if the plugin fails to load, show a toast or console message telling the user what happened and how to fix it

Option 3 seems most robust — there's no reason memory storage should fail just because an image library for the web UI can't load.

### Environment

- **OS:** macOS (Apple Silicon / ARM64)
- **OpenCode plugin version:** opencode-mem 2.13.0
- **Node.js runtime:** managed by OpenCode

leiverkus:
Retry of #90 on the v4 line, per your guidance in #126. You noted the revert (`8fb0836`) was specific to `@huggingface/transformers@4.0.1`'s native ONNX runtime crashing under **Windows + Bun**, not a permanent rejection — so this targets **v4.2.0** (newer onnxruntime-node, `sharp@^0.34.5`) and puts Windows front-and-center in CI.

## Why

`@xenova/transformers@2.17.2` is frozen and pins `sharp@^0.32.0`, which builds its native binary via a **postinstall script**. OpenCode installs plugins with lifecycle scripts skipped → on macOS ARM64 the sharp binary is never created and the plugin throws at load (#94, #97). `@huggingface/transformers@4.2.0` pins `sharp@^0.34.5`, whose binaries ship as prebuilt **`@img/*` optional packages (no install script)** → installs cleanly under a script-skipping installer.

## Changes (minimal — mirrors #90)

- `package.json`: `@xenova/transformers@^2.17.2` → `@huggingface/transformers@^4.2.0`
- `src/services/embedding.ts`: package specifier + types only (`pipeline`/`env` API identical). **Lazy dynamic import preserved** — specifier still built from a non-literal array so the plugin-loader bundle never eagerly traverses the embedding stack.
- guard tests flipped to expect the new backend (still forbidding the old)
- `scripts/verify-embedding-backend.mjs`: reproducible feature-extraction smoke (EN + non-EN, checks dims + L2 norm)
- `.github/workflows/embedding-backend.yml`: **CI matrix ubuntu / macOS / windows**, each running `bun install --ignore-scripts` → native-binary assertion → build → `bun test` → embedding smoke under **both Bun and Node**

## Your requested checks — mapped

| Check | Where |
|---|---|
| keep transformers lazy | preserved; `plugin-bundle-boundary` test enforces it |
| `bun install --ignore-scripts` | CI step + native-binary assertion |
| plugin import/load under cache behavior | `bun run build` + `bun test` (143 pass) |
| small local `feature-extraction` call | `verify-embedding-backend.mjs` in CI |
| macOS ARM64 | verified locally (bun + node) + `macos-latest` in CI |
| **Windows + Bun** | **`windows-latest` in CI — the gate that matters given the 4.0.1 history** |

## Honest scope

I can only verify macOS ARM64 on my own hardware. The Windows and Linux evidence comes from the **CI matrix in this PR** — reproducible, re-runnable by you, not a "trust me" claim. **If the `windows-latest` job is green, the v4.0.1-era regression is resolved.** If it's red, the PR correctly stays blocked and we've learned it cheaply, without another revert.

One caveat I want to be transparent about: CI exercises standalone Bun/Node, not OpenCode's exact plugin-cache sidecar. It covers the ONNX-runtime-load crash that motivated the revert, but a real opencode-on-Windows run would be the final 5%. Happy to coordinate that with you/@NaNomicon if you'd like before merging.

Local results (macOS 26.3, arm64): `bun test` 143/0, typecheck clean, build clean, embedding smoke PASS under bun + node.

Closes #94, closes #97. Discussion: #126.

### DIFF
--- a/.github/workflows/embedding-backend.yml
+++ b/.github/workflows/embedding-backend.yml
@@ -47,11 +47,13 @@ jobs:
           found=$(find node_modules/onnxruntime-node/bin -name '*.node' | head -1)
           test -n "$found" && echo "onnxruntime binding: $found"
 
-      - name: Build
-        run: bun run build
-
-      - name: Test suite
-        run: bun test
+      # This workflow deliberately does NOT run `bun run build` or the full
+      # `bun test` suite. Those have pre-existing cross-platform issues unrelated
+      # to the embedding backend (the build script uses `cp -r`, which fails on
+      # Windows; one config test asserts a macOS-shaped storagePath). They belong
+      # in a general CI workflow, not here. This job's single purpose is to prove
+      # the @huggingface/transformers embedding backend installs and runs across
+      # platforms, which needs neither the built plugin nor the unit tests.
 
       # The revert-critical check: load the ONNX runtime and run a real
       # feature-extraction embedding under Bun on this platform.
--- a/.github/workflows/embedding-backend.yml
+++ b/.github/workflows/embedding-backend.yml
@@ -1,0 +1,68 @@
+name: Embedding Backend Verification
+
+# Verifies the local @huggingface/transformers embedding backend installs and
+# runs across platforms — specifically that the native ONNX runtime + prebuilt
+# sharp (@img/*) load under a script-skipping install on macOS, Linux and
+# Windows. The migration off @xenova/transformers was previously reverted
+# (8fb0836) due to native ONNX crashes under Windows + Bun, so Windows is a
+# first-class target here, not an afterthought.
+
+on:
+  pull_request:
+    paths:
+      - "package.json"
+      - "bun.lock"
+      - "src/services/embedding.ts"
+      - "scripts/verify-embedding-backend.mjs"
+      - ".github/workflows/embedding-backend.yml"
+  workflow_dispatch:
+
+jobs:
+  verify:
+    name: ${{ matrix.os }} / bun
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+
+      # Mimic OpenCode's plugin install, which does not run lifecycle scripts.
+      # The whole point is that prebuilt @img/sharp-* and onnxruntime-node
+      # binaries must arrive without a postinstall build step.
+      - name: Install (no scripts — mimics OpenCode plugin install)
+        run: bun install --ignore-scripts
+
+      - name: Assert native binaries present without postinstall
+        shell: bash
+        run: |
+          set -e
+          echo "Checking prebuilt sharp (@img) ..."
+          ls node_modules/@img/sharp-*/lib/*.node
+          echo "Checking onnxruntime-node native binding ..."
+          # path varies by platform/napi version — just require at least one .node
+          found=$(find node_modules/onnxruntime-node/bin -name '*.node' | head -1)
+          test -n "$found" && echo "onnxruntime binding: $found"
+
+      - name: Build
+        run: bun run build
+
+      - name: Test suite
+        run: bun test
+
+      # The revert-critical check: load the ONNX runtime and run a real
+      # feature-extraction embedding under Bun on this platform.
+      - name: Embedding smoke (Bun)
+        run: bun scripts/verify-embedding-backend.mjs
+
+      # Node path too — OpenCode's plugin sidecar is Node-based in 1.15.x.
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "24"
+
+      - name: Embedding smoke (Node)
+        run: node scripts/verify-embedding-backend.mjs
--- a/bun.lock
+++ b/bun.lock
@@ -5,9 +5,9 @@
     "": {
       "name": "opencode-plugin",
       "dependencies": {
+        "@huggingface/transformers": "^4.2.0",
         "@opencode-ai/plugin": "^1.3.0",
         "@opencode-ai/sdk": "^1.3.0",
-        "@xenova/transformers": "^2.17.2",
         "franc-min": "^6.2.0",
         "iso-639-3": "^3.0.1",
         "usearch": "^2.21.4",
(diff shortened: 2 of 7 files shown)


leiverkus:
## CI matrix is green — including the Windows gate ✅

| Platform | install --ignore-scripts | sharp prebuilt | embedding smoke (Bun) | embedding smoke (Node) |
|---|---|---|---|---|
| ubuntu-latest | ✅ | `@img/sharp-linux-x64` | ✅ | ✅ |
| macos-latest | ✅ | `@img/sharp-darwin-arm64` | ✅ | ✅ |
| **windows-latest** | ✅ | `@img/sharp-win32-x64` | ✅ | ✅ |

The revert-critical evidence, straight from the `windows-latest` log:

```
[verify-embedding] runtime=bun platform=win32 arch=x64
[verify-embedding] loading feature-extraction pipeline for Xenova/all-MiniLM-L6-v2 ...
[verify-embedding] ok: "Hello world..." -> 384 dims, L2=1.0000
[verify-embedding] ok: "Hallo Welt..." -> 384 dims, L2=1.0000
[verify-embedding] PASS — ONNX runtime loaded and embeddings produced.
```

So the `@4.0.1`-era native-ONNX crash under Windows + Bun does **not** reproduce on `@4.2.0`. That was the gate you cared about.

## Heads-up: two PRE-EXISTING cross-platform issues my first CI run surfaced (NOT in this PR's scope)

My initial workflow also ran `bun run build` + full `bun test`, which failed on non-macOS — but **not** because of the migration:

1. **`bun run build` fails on Windows** — the build script is `bunx tsc && mkdir -p dist/web && cp -r src/web/*`; Windows errors with `cp: illegal option -- r`. `tsc` itself compiled fine. This also affects `release.yml`, which runs `bun run build`.
2. **One config test fails on Linux** — `config > CONFIG defaults > should have a storagePath containing .opencode-mem` asserts a macOS-shaped path.

I scoped this workflow down to just the embedding-backend proof so these don't block it. Happy to send a separate small PR fixing both (e.g. a tiny cross-platform copy step + making the storagePath test platform-aware) if you'd like — they're worth fixing since they mean the project currently only builds/tests cleanly on macOS.

## Honest residual gap

CI exercises standalone Bun/Node, not OpenCode's exact plugin-cache sidecar. It proves the ONNX-runtime-load crash is gone cross-platform; a real opencode-on-Windows memory write would be the final 5%. I run opencode on macOS only, so I can't do that one — but given the matrix is green, the risk surface is small. Your call on whether you want that extra confirmation before merging.

PR #101, opened 2026-04-23
refactor(ai): use opencode v2 SDK session.prompt for structured output

StumHuang:
## Summary

Replaces the auth.json-reading + manual OAuth refresh + direct provider HTTPS flow with opencode's v2 client SDK. Per call we open a transient session, prompt with `format: json_schema`, then delete the session.

## Why

- opencode already owns the user's auth, token refresh, and provider routing for every connected provider; reimplementing that here was fragile (broke whenever opencode tweaked auth.json layout) and limited us to anthropic + openai.
- Going through `session.prompt` unlocks every provider opencode supports out of the box, including **github-copilot personal/business plans**, OAuth flows like Claude Pro/Max, and any custom provider users wire up.

## What changed

- **`src/services/ai/opencode-provider.ts`**: 363 → 147 lines. Drop auth.json parser, OAuthFetch with token refresh, and AI-SDK Anthropic/OpenAI clients. Add `createV2Client` + `getV2Client` + new `generateStructuredOutput` signature taking `{ client, providerID, modelID, ... }`.
- **`src/index.ts`**: wire `setV2Client(createV2Client(ctx.serverUrl))` at plugin init; drop now-unused `setStatePath(client.path.get())` call.
- **`src/services/auto-capture.ts`** and **`user-memory-learning.ts`**: migrate to the new `generateStructuredOutput` signature; fail fast if v2 client is not yet initialized.
- **`tests/opencode-provider.test.ts`**: rewrite to mock `OpencodeClient` and cover session lifecycle, schema serialization, and error paths.
- **`package.json` + `bun.lock`**: drop `@ai-sdk/anthropic`, `@ai-sdk/openai`, `ai`.
- **`src/config.ts` + `README.md`**: document `github-copilot` as a first-class supported provider.

## Verification

- `bun run typecheck` — clean
- `bun test tests/opencode-provider.test.ts` — **11 pass / 0 fail** (27 expect calls)
- Pre-commit hook (typecheck + prettier) passed

## Net diff

+468 / −437 across 9 files. The +s are mostly new test coverage for the v2 client mock; the production module shrinks substantially.

## Note on ordering

Independent of #100 in scope, but the diffs touch overlapping lines in `src/index.ts` (warmup wiring). Recommend merging #100 first; happy to rebase this on top once #100 lands.

PR #140, opened 2026-06-26
fix(ai): bypass v2 SDK session methods for structured output (#110)

### LINKED ISSUE #110: Auto-capture broken: SDK v1.14.48 has no session.create method
## Bug

Auto-capture fails on every idle event with:

```
Error: opencode-mem: session.create returned no session id; cannot generate structured output
```

## Environment

- opencode-mem: v2.14.1
- OpenCode: v1.14.48
- @opencode-ai/sdk: v1.14.48
- OS: Windows 11
- Provider: github-copilot

## Root Cause

PR #101 (merged May 9) replaced the old auth.json/direct-API flow with `client.session.create()` via the v2 SDK. However, the `Session` class in `@opencode-ai/sdk` v1.14.48 only exposes a `list()` method — there is no `create()`, `prompt()`, or `delete()` method on the session object.

In `sdk/dist/v2/gen/sdk.gen.js`, the `Session` class (line 382) only has:
```js
export class Session extends HeyApiClient {
    list(parameters, options) { ... } // GET /experimental/session
}
```

So `client.session.create()` returns `undefined`, and `created?.data?.id` is `undefined`, triggering the error.

## Impact

- Auto-capture is **100% broken** — no memories are ever saved automatically
- The manual `memory add` tool still works fine
- The `chat.message` hook (memory injection) still works fine

## Log Evidence

Every single idle event since installation fails:

```
--- Session started: 2026-05-11T08:48:59.896Z ---
[2026-05-11T08:48:59.897Z] Idle processing error: {"error":"Error: opencode-mem: session.create returned no session id; cannot generate structured output"}
[2026-05-11T08:50:04.949Z] Idle processing error: {"error":"Error: opencode-mem: session.create returned no session id; cannot generate structured output"}
... (74+ failures logged)
```

## Suggested Fix

Either:
1. Wait for the SDK to publish the `session.create`/`prompt`/`delete` methods, or
2. Fall back to raw `fetch` against the OpenCode HTTP API (`POST /session`, `POST /session/{id}/prompt`, `DELETE /session/{id}`) which does exist on the server

## Workaround

None currently. The v2.13.0 approach (auth.json + direct Anthropic/OpenAI API) doesn't support `github-copilot` provider either.

kumar-shivang:
## Summary

PR #101 switched structured-output generation in `opencode-provider.ts` to
`client.session.create / client.session.prompt / client.session.delete` on
`@opencode-ai/sdk` v2. On `@opencode-ai/sdk@1.14.48` this regresses auto-capture
100% of the time: the only `Session` class the published bundle exposes at
`client.session` has shifted across releases (in v1.14.48 it is a thin
experimental class with only `list()`; the real `create/prompt/delete` live
on a differently-named class reachable via a different property path), so the
`client.session.create(...)` call returns undefined and
`created?.data?.id` is `undefined`, triggering:

```
Error: opencode-mem: session.create returned no session id; cannot generate structured output
```

on every idle event. See #110.

## Fix

Stop using the SDK's session methods for these three calls. Hit the
documented server endpoints directly with raw `fetch`:

| action      | endpoint                                      |
| ----------- | --------------------------------------------- |
| create      | `POST /session`                               |
| prompt      | `POST /session/{id}/message`                  |
| delete      | `DELETE /session/{id}`                        |

These paths are stable across server versions and match exactly what
`tests/opencode-provider.test.ts` already mocks, so the existing test suite
is preserved.

The base URL is captured when `createV2Client(serverUrl)` is called and
cached in a module-level `_v2BaseUrl`; the `client` argument on
`StructuredOutputOptions` is retained for API compatibility but is no
longer used. The `setV2Client` / `getV2Client` / `createV2Client` exports
are preserved so the rest of the codebase (e.g. `src/index.ts`,
`src/services/auto-capture.ts`, `src/services/user-memory-learning.ts`) is
unaffected.

## Tests

- `bun run typecheck` — clean
- `bun test` — 162 pass / 0 fail (was 154 baseline; +8 new tests)
- `bunx prettier --check` on changed files — clean

New tests in `tests/opencode-provider.test.ts` (the
`generateStructuredOutput regression tests (issue #110)` block) cover what
the previous suite did not:

1. `directory` is forwarded as a percent-encoded query param on all three
   calls (create, prompt, delete).
2. `directory` is omitted (no empty `?directory=`) when not provided.
3. Base URL with a trailing slash is normalized — no `//session` double
   slash.
4. `createV2Client` accepts a `URL` object (not just a string).
5. Non-2xx response from `POST /session` surfaces a clear error and skips
   prompt/delete.
6. Non-2xx response from `POST /session/{id}/message` surfaces the error
   and still runs the best-effort `DELETE`.
7. Network error (`fetch` rejection) from `POST /session` propagates and
   skips prompt/delete.
8. Public API surface (`generateStructuredOutput`, `createV2Client`) is
   preserved.

## Notes

- The previous implementation happened to pass the existing tests because
  they stub `globalThis.fetch`, and the v2 SDK's underlying hey-api client
  ultimately calls `globalThis.fetch`. So the test suite stayed green even
  though the production call path was broken — the fix here is verified
  by reproducing the same `globalThis.fetch` mock in new tests.
- The issue's suggested fix (option 2: fall back to raw `fetch`) is what
  this PR implements. Option 1 (wait for the SDK to publish the methods)
  is unnecessary once we talk to the server directly.

## Refs

- Closes #110
- Related: #101 (the PR that introduced the regression)

### DIFF
--- a/src/services/ai/opencode-provider.ts
+++ b/src/services/ai/opencode-provider.ts
@@ -1,13 +1,26 @@
 /**
- * SDK-based structured output via opencode v2 session.prompt.
+ * Structured output via the opencode HTTP server.
  *
- * Replaces the old auth.json/OAuth-juggling flow. Instead of forging requests
- * to provider HTTP endpoints ourselves, we delegate to the running opencode
- * server: it already owns the user's auth (any provider, including
- * github-copilot personal/business), token refresh, and provider routing.
+ * Replaces the older auth.json/OAuth-juggling flow. Instead of forging
+ * requests to provider HTTP endpoints ourselves, we delegate to the
+ * running opencode server: it already owns the user's auth (any provider,
+ * including github-copilot personal/business), token refresh, and provider
+ * routing.
  *
- * Per call we create a transient session, prompt with a JSON schema, then
- * delete the session so it does not pollute the user's TUI session list.
+ * Per call we create a transient session, prompt it with a JSON schema,
+ * then delete the session so it does not pollute the user's TUI session
+ * list.
+ *
+ * We intentionally bypass the `@opencode-ai/sdk` client for these three
+ * endpoints. Issue #110 showed that relying on `client.session.create` /
+ * `client.session.prompt` / `client.session.delete` is brittle: the SDK
+ * class layout has shifted across releases (e.g. v1.14.48's `Session` only
+ * exposes `list()` in some builds, with the real methods living on a
+ * renamed `Session2` reachable via a different property path). Going
+ * straight to `fetch` against the documented server endpoints
+ * (`POST /session`, `POST /session/{id}/message`, `DELETE /session/{id}`)
+ * makes us resilient to those SDK churns and lets us test the wire
+ * protocol directly with a `globalThis.fetch` stub.
  */
 
 import type { z } from "zod";
@@ -15,6 +28,7 @@ import { createOpencodeClient, type OpencodeClient } from "@opencode-ai/sdk/v2/c
 
 let _connectedProviders: Set<string> = new Set();
 let _v2Client: OpencodeClient | undefined;
+let _v2BaseUrl: string | undefined;
 
 export function setConnectedProviders(providers: string[]): void {
   _connectedProviders = new Set(providers);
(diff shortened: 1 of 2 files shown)


prihuda:
is there any news for this to merged ?
because of this problem i need to pinned on v2.13.0

davischw:
@tickernelz Can we get this merged somehow?

tickernelz:

PR #143, opened 2026-06-28
Clarify config and auto-capture docs

### LINKED ISSUE #105: Documentation needs updating
1. The config does not match anything like the template in code.
2. There is no documentation about auto capture needing to be connected to a third party for it to work.
3. There is no documentation about libraries needing to be installed prior to using this plugin. Ran into that issue yesterday.

x-Spartacus:
Closes #105

- Document first-use/source install prerequisites for embeddings and dependencies
- Clarify that the README config is a trimmed example and the full commented template is generated on first startup
- Document auto-capture provider requirements, opencodeProvider precedence, manual provider modes, proxy/tool-call limitations, and temperature troubleshooting

Tested:
- npx prettier --check README.md
- npx bun run typecheck

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,11 +30,14 @@ This plugin uses `USearch` for preferred in-memory vector indexing with automati
 
 - Bun
 - Standard OpenCode plugin environment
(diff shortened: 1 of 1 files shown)


tickernelz:

PR #146, opened 2026-06-28
fix(auto-capture): reuse injected opencode client

### LINKED ISSUE #145: Auto-capture uses unreachable ctx.serverUrl in no-port OpenCode sessions
## Summary

`opencode-mem` auto-capture can fail with:

```text
opencode-mem: session.create returned no session id; cannot generate structured output
```

when OpenCode is started without an explicit `--port`.

## Root cause observed

In a no-port OpenCode session, the plugin received:

```text
ctx.serverUrl = http://localhost:4096/
```

but no OpenCode REST API was listening there. The structured-output path creates a fresh SDK client from `ctx.serverUrl` and calls `client.session.create()`, so the request fails before a session id is returned.

A local diagnostic patch showed the underlying SDK result was an error result, not a successful session response:

```json
{
  "topKeys": ["error", "request", "response"],
  "hasDataId": false,
  "hasError": true,
  "requestPath": "/session",
  "errorMessage": "Unable to connect. Is the computer able to access the url?"
}
```

The current user-facing error hides this underlying connection failure because `generateStructuredOutput()` only checks `created?.data?.id`.

## Control case

When starting OpenCode with an explicit port, for example `--port 61234`, the plugin received:

```text
ctx.serverUrl = http://127.0.0.1:61234/
```

and `session.create` succeeded with `created.data.id` present.

## Expected behavior

Auto-capture should not depend on an unreachable default `ctx.serverUrl` when OpenCode already injects a working `ctx.client`. The structured-output path should reuse the injected client or otherwise surface the real connection error.

## Environment

- OpenCode: 1.17.11
- opencode-mem: 2.17.4
- Node: 22.22.1
- OS: Linux

## Proposed fix

Reuse the injected `ctx.client` through a small adapter for `session.create`, `session.prompt`, and `session.delete`, instead of creating a second SDK client from `ctx.serverUrl`. Also include the underlying SDK error details when `session.create` returns no id.

I have a PR ready with a regression test.

huilang021x:
## Summary

- What changed: Auto-capture now builds its structured-output session client from the injected OpenCode `ctx.client` instead of creating a second client from `ctx.serverUrl`. `session.create` failures now include underlying status/error details when available.
- Why: In no-port OpenCode sessions, `ctx.serverUrl` can be `http://localhost:4096/` even when no REST API is listening there, causing `session.create returned no session id` during auto-capture.
- Current phase: Ready for review. Closes #145.

## Scope

- In scope: Structured-output auto-capture client initialization and regression coverage for the injected-client adapter.
- Out of scope: Provider/model timeout tuning, storage/indexing behavior, web UI behavior.
- Risk level: Low/medium; changes the internal client used for auto-capture session create/prompt/delete while preserving the existing request shapes.

## Deployment Targets

- Impacted services: `opencode-mem` plugin auto-capture and user-profile structured-output paths that use the shared opencode provider client.
- Restart/deploy needed: Users need to restart OpenCode after upgrading the plugin.
- Target environment: OpenCode plugin runtime.
- Rollback note: Revert this commit to restore the previous `ctx.serverUrl` client path.

## Acceptance Plan

- Acceptance criteria: Auto-capture should not fail with no session id solely because `ctx.serverUrl` points to an unreachable default port; structured-output calls should use the injected client and still pass `format`, `system`, `model`, and `parts` correctly.
- Evidence to collect: Regression test verifies `createStructuredOutputClient()` maps v2-style calls onto the injected client without opening a new serverUrl client.
- Required checks: Typecheck, build, formatter check, targeted provider tests, full test suite.
- Manual verification needed: Optional validation in a no-port OpenCode runtime after publishing.

## Validation

- Commands run:
  - `bun test tests/opencode-provider.test.ts`
  - `bun run format:check`
  - `bun run typecheck`
  - `bun run build`
  - `bun test`
- Result: All passed after formatting.
- Evidence links/log snippets:
  - Targeted test: 12 pass, 0 fail
  - Full test suite: 157 pass, 0 fail

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -45,11 +45,13 @@ export const OpenCodeMemPlugin: Plugin = async (ctx: PluginInput) => {
 
   (async () => {
     try {
-      const { setConnectedProviders, setV2Client, createV2Client } =
+      const { createStructuredOutputClient, setConnectedProviders, setV2Client } =
         await import("./services/ai/opencode-provider.js");
-      setV2Client(createV2Client(ctx.serverUrl));
+      setConnectedProviders([]);
+      setV2Client(undefined);
       const providerResult = await ctx.client.provider.list();
       if (providerResult.data?.connected) {
+        setV2Client(createStructuredOutputClient(ctx.client));
         setConnectedProviders(providerResult.data.connected);
       }
     } catch (error) {
--- a/src/services/ai/opencode-provider.ts
+++ b/src/services/ai/opencode-provider.ts
@@ -11,10 +11,42 @@
  */
 
 import type { z } from "zod";
-import { createOpencodeClient, type OpencodeClient } from "@opencode-ai/sdk/v2/client";
+import { createOpencodeClient } from "@opencode-ai/sdk/v2/client";
+
+type SessionCreateResult = {
+  data?: { id?: string };
+  id?: string;
+  error?: { message?: string; data?: { message?: string } };
+  response?: { status?: number; statusText?: string };
+};
+
+type SessionPromptResult = {
+  data?: {
+    info?: {
+      structured?: unknown;
+      error?: { name: string; data?: { message?: string } };
+    };
+  };
+};
+
+export interface StructuredOutputClient {
+  session: {
+    create(input: Record<string, unknown>): Promise<SessionCreateResult>;
+    prompt(input: Record<string, unknown>): Promise<SessionPromptResult>;
+    delete(input: Record<string, unknown>): Promise<unknown>;
+  };
+}
+
+type InjectedOpencodeClient = {
+  session: {
+    create(options: unknown): Promise<unknown>;
+    prompt(options: unknown): Promise<unknown>;
+    delete(options: unknown): Promise<unknown>;
+  };
+};
 
 let _connectedProviders: Set<string> = new Set();
-let _v2Client: OpencodeClient | undefined;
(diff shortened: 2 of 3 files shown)


tickernelz:
I tried to merge this PR into the latest `main` as part of a batch release, but it currently has merge conflicts in:

- `src/services/ai/opencode-provider.ts`

Several structured-output / opencode-provider changes landed in the latest release, which is what this now collides with. Please rebase/merge the latest `main` and resolve the conflicts, and I'll merge it afterward.

prihuda:
i need this because on 2.18.0 the problem still persist

GraDea:
The goal is right: structured auto-capture should reuse OpenCode’s running host context instead of depending on a separately reachable localhost server.

I’m going to hold/close this in favor of #161, because this PR routes back through `ctx.client.session.create/prompt/delete`, while #140 intentionally moved away from those SDK/session helper methods due SDK shape brittleness.

The replacement approach keeps #140’s raw endpoint direction, but uses the host SDK transport (`fetch`/`baseUrl`) underneath it. It also covers the follow-up issues we found in runtime testing:

- injected host fetch needs `Request` object calls
- `noReply: true` suppresses structured output in current OpenCode builds
- `info.structured_output` must be supported
- diagnostics need redaction
- stale host transport state must be reset between plugin initializations

Closing as superseded by #161, not rejected.

PR #147, opened 2026-06-28
Avoid forced process exit on shutdown

### LINKED ISSUE #114: [Bug] Bun NAPI crash on TUI exit: process.exit() in SIGINT handler conflicts with onnxruntime-node native cleanup
## Description

When using `opencode-mem` plugin with Bun runtime, exiting the TUI via Ctrl+C consistently crashes Bun with:

```
panic: NAPI FATAL ERROR: Error::New napi_create_error
```

## Root Cause

The plugin registers a SIGINT/SIGTERM handler in `dist/index.js` that calls `process.exit()` after cleanup. When exiting opencode TUI:

1. Plugin's SIGINT handler fires, calls `memoryClient.close()` then `process.exit(0)`
2. `onnxruntime-node` (used via `@xenova/transformers` for local embeddings) has an `InferenceSession` loaded in a worker thread
3. `process.exit()` triggers process termination — Bun starts tearing down the NAPI environment
4. The ONNX Runtime GC finalizer (`InferenceSessionWrap::FinalizeCallback`) tries to destroy the session
5. During destruction, `Napi::Error::New(napi_env__*)` is called, but the `napi_env` is already partially destroyed
6. Bun panics with `NAPI FATAL ERROR: Error::New napi_create_error`

## Crash Stack (from macOS crash report)

```
InferenceSessionWrap::FinalizeCallback()
  → InferenceSessionWrap::~InferenceSessionWrap()
    → Napi::ObjectWrap<InferenceSessionWrap>::~ObjectWrap()
      → Napi::Error::New(napi_env__*)
        → Napi::Error::Fatal() → Bun panic
```

## Environment

- Bun: 1.3.14
- opencode-mem: 2.14.3
- opencode: 1.15.10
- macOS 15.5 (Apple Silicon arm64)
- onnxruntime-node: 1.14.0

## Suggested Fix

Remove `process.exit()` from the `shutdownHandler` — let the host process manage lifecycle:

```js
const shutdownHandler = async () => {
    try {
        if (webServer) await webServer.stop();
        memoryClient.close();
        // DO NOT call process.exit()
    } catch (error) {
        log("Shutdown error", { error: String(error) });
    }
};
```

## Related

See also #97, #88 for other NAPI/native addon issues on arm64.

x-Spartacus:
Closes #114

- Remove `process.exit()` calls from the SIGINT/SIGTERM shutdown handler
- Preserve cleanup and error logging
- Clear pending idle auto-capture timeout during cleanup
- Set `process.exitCode = 1` on cleanup failure without forcing host process teardown
- Add regression coverage so `process.exit()` is not reintroduced

Tested:
- npx bun test tests/plugin-shutdown.test.ts tests/plugin-loader-contract.test.ts
- npx bun run typecheck
- npx bun run build
- npx bun test

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -129,6 +129,10 @@ export const OpenCodeMemPlugin: Plugin = async (ctx: PluginInput) => {
   }
 
   const cleanupPlugin = async () => {
+    if (idleTimeout) {
+      clearTimeout(idleTimeout);
+      idleTimeout = null;
+    }
     if (webServer) await webServer.stop();
     if (memoryClient) memoryClient.close();
   };
@@ -136,10 +140,9 @@ export const OpenCodeMemPlugin: Plugin = async (ctx: PluginInput) => {
   const shutdownHandler = async () => {
     try {
       await cleanupPlugin();
-      process.exit(0);
     } catch (error) {
       log("Shutdown error", { error: String(error) });
-      process.exit(1);
+      process.exitCode = 1;
     }
   };
 
--- a/tests/plugin-shutdown.test.ts
+++ b/tests/plugin-shutdown.test.ts
@@ -1,0 +1,17 @@
+import { describe, expect, it } from "bun:test";
+import { readFileSync } from "node:fs";
+
+describe("plugin shutdown", () => {
+  it("does not force host process exit from signal handlers", () => {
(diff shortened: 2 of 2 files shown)


tickernelz:

PR #148, opened 2026-06-29
feat(tags): resolve project root from a workspace marker file

black-desk:
## Problem

Project identity is derived from the enclosing git repository (git-common-dir, then remote origin URL, then the directory path), which splits memory into one shard per physical git repo. That is wrong for multi-repo workspaces managed by orchestrators such as Google `repo` that lay out many git repositories under a single workspace root: each sub-repository gets its own isolated memory store instead of sharing one project-wide store.

## Solution: directory-driven workspace marker

Drop an empty `.opencode-mem-project` marker file at the workspace root. Every session started anywhere underneath resolves onto that root.

The marker is looked up by walking up from the working directory that **every code path already passes in** — the plugin's `ctx.directory`, the web API's `process.cwd()`. Identity is therefore bound to *where the session runs*, not to ambient process state, so it stays consistent regardless of which process serves a given request.

```
my-workspace/
├── .opencode-mem-project   ← workspace root
├── kernel/                 (own git repo)
├── userspace/              (own git repo)
└── tools/                  (own git repo)
```

```sh
touch ~/my-workspace/.opencode-mem-project
```

## Behavior

- The walk lives in `getProjectRoot` / `getProjectIdentity` (the lowest-level entry points), so plugin load, auto-capture, user-profile learning, compaction, and the web API all pick it up with no per-callsite changes.
- The marker takes precedence over git detection. When it hits, the underlying sub-repo's git remote is left unset (it would describe only one nested repo and be misleading for the grouped workspace).
- **Without a marker, behaviour is unchanged** — existing git-based identity is identical.
- Innermost marker wins when several are nested on the ancestry.
- The marker lookup runs exactly once per `getProjectTagInfo` call; the git-only fallbacks are factored into private helpers so root and identity derive from a single ancestry walk.

## Testing

New cases in `tests/project-scope.test.ts` cover: sibling git repos collapse onto one marker root; deep nested paths resolve up to the marker; the marker wins over an inner git repo and drops its remote; `findMarkerProjectRoot` null / ancestor / self cases; plus the backward-compatible no-marker behaviour (the existing worktree and nested-path tests still pass).

`tsc --noEmit`, `prettier`, and `bun run build` are clean. The pre-commit hook (typecheck + lint-staged) passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,46 @@ Configure at `~/.config/opencode/opencode-mem.jsonc`:
 - `scope: "all-projects"`: query `search` / `list` across all project shards.
 - `memory.defaultScope` sets the default query scope when no explicit scope is provided.
 
+### Sharing One Project Memory Across Nested Repos
+
+By default a project is identified by its enclosing git repository, so every
+physical git repo gets its own isolated memory store. That is wrong for
+multi-repo workspaces — trees managed by Google [`repo`](https://gerrit.googlesource.com/git-repo/+/HEAD/Docs/manual-repo.md),
+monorepos, or any layout where several nested git repositories belong to one
+logical project — because each sub-repository would be siloed.
+
+Drop an empty **`.opencode-mem-project`** marker file at the workspace root:
+
+```
+my-workspace/
+├── .opencode-mem-project   ← workspace root
+├── kernel/                 (own git repo)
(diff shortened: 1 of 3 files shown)


tickernelz:

PR #151, opened 2026-07-02
feat: `opencodeModel: "inherit"` — capture with the session's own model

anagnorisis2peripeteia:
## Motivation

Pinned `opencodeModel` ids rot: when a provider renames a model (or a local endpoint changes its served model id), auto-capture silently fails at `session.create` on every prompt until someone notices the log spam. That is how I found this — capture had been broken for days on a stale local-model id. Separately, users who switch models during the day may prefer each conversation to be captured by the model that actually produced it.

## What this does

`"opencodeModel": "inherit"` makes auto-capture use the exact provider/model opencode resolved for the captured prompt:

- a `chat.params` hook records `input.model.providerID` / `input.model.id` onto the prompt's existing `user_prompts` row (two new nullable columns, added with the same duplicate-column-guarded `ALTER TABLE` pattern as `capture_attempts`), keyed by `message.id` — the same key capture already uses, so the mapping is exact rather than inferred from transcript position;
- at capture time, `"inherit"` resolves provider/model from the stored row; the existing `isProviderConnected` guard runs against the resolved provider;
- if a prompt has no recorded model (predates the upgrade, or the provider path didn't invoke `chat.params`), capture throws and the existing retry/cap machinery handles it — it never guesses.

Pinned configuration behaves exactly as before; the hook returns immediately unless `"inherit"` is configured.

## Notes for review

- `opencodeProvider` must still be set (the existing `opencodeProvider && opencodeModel` gate is untouched for diff minimality); with `"inherit"` its value is only used as a gate. Happy to change the gate if you prefer.
- On opencode v1.17.11 I observed `chat.params` firing reliably for models served through the standard providers, but not for every custom npm-provider path — the hard-error fallback covers those prompts. Worth knowing for docs/support.

## Testing

- `tsc --noEmit` and Prettier clean; migration verified against an existing populated `user-prompts.db` (old rows read back with `null` model fields).
- Live-tested against opencode v1.17.11: sessions on a local llama.cpp-served model recorded `provider_id`/`model_id` correctly on their prompt rows, and capture consumed the recorded model (verified via capture-session behavior/logs).
- Rows without recorded models fail capture with the new explicit error and are retried/capped by the existing machinery.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,8 @@ Supported providers: any provider listed by `opencode providers list` (e.g. `ant
 
 If `opencodeProvider` and `opencodeModel` are set, they take precedence over the manual `memoryProvider` settings below.
 
+**Follow the session model:** set `"opencodeModel": "inherit"` to capture each prompt with the exact provider/model opencode used to answer it (recorded per message via the `chat.params` hook). This is useful if you switch models often or use several local backends; captures then always use the model that produced the conversation, and pinned ids can’t become stale. `opencodeProvider` is still required as the normal config gate, and is used if a prompt has no recorded model.
+
 **Fallback:** Manual API configuration (if not using opencodeProvider):
 
 ```jsonc
--- a/src/index.ts
+++ b/src/index.ts
@@ -309,6 +309,16 @@ export const OpenCodeMemPlugin: Plugin = async (ctx: PluginInput) => {
       }
     },
 
+    "chat.params": async (input) => {
+      if (!isConfigured() || CONFIG.opencodeModel !== "inherit") return;
+
+      try {
+        userPromptManager.setPromptModel(input.message.id, input.model.providerID, input.model.id);
+      } catch (error) {
+        log("chat.params: ERROR", { error: String(error) });
+      }
+    },
+
     tool: {
       memory: tool({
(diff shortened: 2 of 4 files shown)


tickernelz:
I tried to merge this PR into the latest `main` as part of a batch release, but it currently has merge conflicts in:

- `README.md`

Please rebase/merge the latest `main` and resolve the conflicts, and I'll merge it in the next release.

anagnorisis2peripeteia:
I’ve rebased this PR onto latest main and resolved the README.md conflict. Branch is updated and force-pushed at commit 2286d24, so this should now be ready for merge.

anagnorisis2peripeteia:
@tickernelz , should be good

tickernelz:
I tried to merge this PR into the latest `main` as part of a batch release, but it currently has merge conflicts in:

- `src/services/auto-capture.ts`

Please rebase/merge the latest `main` and resolve the conflicts, and I'll merge it in the next release.

anagnorisis2peripeteia:
> I tried to merge this PR into the latest `main` as part of a batch release, but it currently has merge conflicts in:
> 
> * `src/services/auto-capture.ts`
> 
> Please rebase/merge the latest `main` and resolve the conflicts, and I'll merge it in the next release.

@tickernelz done

tickernelz:
Merged into `main` and released in v2.19.1. Verified on the fully-merged tree: `bun run typecheck`, `bun run build`, and `bun test` (197 pass / 0 fail) all green. Thanks for rebasing and resolving the earlier conflict!

PR #155, opened 2026-07-05
Refactor/auto capture provider readiness

GraDea:
## Summary

This refactors auto-capture provider readiness so invalid configuration is detected and reported from the config layer instead of being re-derived inside the auto-capture runtime path.

Changes:
- add `CONFIG.autoCaptureProviderStatus`
- add `getAutoCaptureProviderStatus()` with explicit readiness mode/issues
- report all missing/invalid provider settings independently
- move invalid auto-capture config logging to plugin startup
- make `performAutoCapture()` silently skip when provider config is not ready
- improve manual provider config errors to list missing fields
- require all three manual provider fields: `memoryModel`, `memoryApiUrl`, and `memoryApiKey`
- keep placeholder key detection for `sk-...`, `sk-ant-...`, `gsk_...`, etc.

## Why

PR #154 fixed placeholder API keys reaching provider setup, but the readiness logic still lived partly in `auto-capture.ts` and partly in provider config validation.

This follow-up makes the intended state explicit:

- OpenCode provider mode is ready when `opencodeProvider` and `opencodeModel` are configured.
- Manual provider mode is ready when `memoryModel`, `memoryApiUrl`, and a non-placeholder `memoryApiKey` are configured.
- Invalid states expose every config issue at once instead of failing one check at a time.

This also removes the runtime “log once” flag from auto-capture. The warning is now logged once naturally during plugin initialization.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -3,7 +3,7 @@ import { join } from "node:path";
 import { homedir } from "node:os";
 import { stripJsoncComments } from "./services/jsonc.js";
 import { resolveSecretValue } from "./services/secret-resolver.js";
-import { allowsMissingApiKey, isPlaceholderApiKey } from "./services/ai/api-key-placeholder.js";
+import { isPlaceholderApiKey } from "./services/ai/api-key-placeholder.js";
 
 const CONFIG_DIR = join(homedir(), ".config", "opencode");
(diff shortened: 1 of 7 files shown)


tickernelz:

PR #156, opened 2026-07-06
fix(sqlite): avoid require in shard deletion

fengjikui:
## Summary
- replace the CommonJS `require("node:fs")` in `deleteShard()` with static ESM `node:fs` imports
- keep shard file cleanup working when the package is loaded as an ES module by OpenCode
- add a plugin-loader contract regression for the built shard-manager output

## Root cause
`opencode-mem` is published as `type: module`, but `deleteShard()` used a local CommonJS `require("node:fs")`. In Node/OpenCode ESM loading, that can throw `require is not defined`, leaving the shard DB file behind after metadata/index cleanup.

## Validation
- `bun install`
- `bun run build`
- `bun test tests/plugin-loader-contract.test.ts`
- `bun run typecheck`
- `bun run format:check`
- `bun test`
- `git diff --check`

Related to #117

### DIFF
--- a/src/services/sqlite/shard-manager.ts
+++ b/src/services/sqlite/shard-manager.ts
@@ -1,6 +1,6 @@
 import { getDatabase } from "./sqlite-bootstrap.js";
 import { join, basename } from "node:path";
-import { existsSync } from "node:fs";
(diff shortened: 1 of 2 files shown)


tickernelz:

PR #160, opened 2026-07-07
fix(auto-capture): capture all idle prompts

### LINKED ISSUE #159: Bug: auto-capture on session.idle only processes the last uncaptured prompt
It looks like when `session.idle` triggers auto-capture, the current logic only processes the latest uncaptured prompt in the session.

If multiple prompts are sent before the session becomes idle, only the last one seems to be captured as memory, while earlier prompts remain uncaptured. This may lead to incomplete memory capture for a single conversation session.

Is this intentional?

If not, should the expected behavior be:
- process all uncaptured prompts in the session
- handle them in chronological order
- keep each memory scoped to its own prompt/response block

fengjikui:
## Summary
- Process every uncaptured prompt for the idle session in chronological order instead of only the latest one.
- Scope each auto-capture summary to the assistant/tool messages before the next user prompt.
- Add regression coverage for multi-prompt idle capture and session-scoped prompt ordering.

Fixes #159

## Validation
- bun test tests/auto-capture.test.ts tests/user-prompt-manager-claim.test.ts
- bun run typecheck
- bun run build
- bun test
- bun run format:check
- git diff --check

### DIFF
--- a/src/services/auto-capture.ts
+++ b/src/services/auto-capture.ts
@@ -3,7 +3,7 @@ import { memoryClient } from "./client.js";
 import { getTags } from "./tags.js";
 import { log } from "./logger.js";
 import { CONFIG } from "../config.js";
-import { userPromptManager } from "./user-prompt/user-prompt-manager.js";
+import { userPromptManager, type UserPrompt } from "./user-prompt/user-prompt-manager.js";
 
 interface ToolCallInfo {
   name: string;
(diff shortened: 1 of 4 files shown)


tickernelz:
Merged into `main` as part of the v2.19.0 batch release. Verified on the fully-merged tree: `bun run typecheck`, `bun run build`, and `bun test` (197 pass / 0 fail) all green. Thanks!

PR #161, opened 2026-07-08
Pr 140 bypass sdk

GraDea:
## Summary

Follow-up to merged #140. PR #140 fixed the first layer of issue #110 by bypassing brittle OpenCode v2 SDK session helper methods and calling the raw session HTTP endpoints directly.

This PR fixes the next runtime layer: in the OpenCode plugin host, those raw HTTP calls still should not rely on plain/global `fetch` assumptions. They need to use OpenCode’s host SDK transport (`fetch`/`baseUrl`) when available.

## What #140 fixed

#140 replaced calls through generated SDK session helpers such as:

- `client.session.create`
- `client.session.prompt`
- `client.session.delete`

with direct calls to the OpenCode server endpoints:

- `POST /session`
- `POST /session/{id}/message`
- `DELETE /session/{id}`

That addressed SDK method-shape incompatibility across OpenCode/SDK versions.

## Why this follow-up is still needed

After #140 was merged, we tested fresh `origin/main` at `6b564a0` through the real plugin surface:

- checked out fresh `main`
- reset to `origin/main`
- confirmed local OpenCode config points to `/home/peshkov/src/opencode-mem`
- confirmed auto-capture config:
  - `autoCaptureEnabled: true`
  - `opencodeProvider: openai`
  - `opencodeModel: gpt-5.5`
- ran `bun install`
- ran `bun run build`
- restarted OpenCode and triggered manual auto-capture

Main still failed at runtime:

```text
Auto-capture warning (attempt 1/3): Unable to connect. Is the computer able to access the url?
Auto-capture warning (attempt 2/3): Unable to connect. Is the computer able to access the url?
Auto-capture final error after 3 attempts: Unable to connect. Is the computer able to access the url?
```

So #140 fixed SDK helper incompatibility, but not the plugin-host transport failure.
Changes in this PR
- Extract OpenCode host SDK fetch/baseUrl from nested plugin client services.
- Use host-injected fetch for structured-output session requests when available.
- Reset stale host fetch before configuring transport on plugin startup.
- Fall back to global fetch with a warning log if host fetch cannot be discovered.
- Call the host fetch with Request objects, matching the SDK fetch contract.
- Preserve structured output generation by intentionally omitting noReply, since noReply suppresses assistant generation and prevents info.structured_output.
- Support both info.structured_output and legacy info.structured.
- Redact sensitive diagnostics from OpenCode HTTP failures:
- strip query params from diagnostic URLs
- redact response bodies
- allowlist only safe assistant error fields
- Skip capturing the plugin’s own structured-summary prompt echo to avoid self-referential memories.
- Add auto-capture outcome logs for skipped and persisted captures.
- Keep dist clean before builds while preserving the current scripts/copy-web-assets.mjs asset-copy path.
- Add regression coverage for:
- nested host fetch/baseUrl extraction
- stale host fetch reset
- fallback warning log
- host fetch usage over global fetch
- Request-based fetch invocation
- no-noReply structured output requests
- structured_output parsing
- diagnostic redaction
- summary prompt filtering
Verification
On this branch:
- bun run typecheck
- bun run format:check
- bun run build && bun test
Result:
198 pass
0 fail
Also manually verified earlier that the fixed branch can persist auto-capture memory:
Auto-capture memory persisted

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     }
   },
   "scripts": {
-    "build": "bunx tsc && bun scripts/copy-web-assets.mjs",
+    "build": "rm -rf dist && bunx tsc && bun scripts/copy-web-assets.mjs",
     "dev": "tsc --watch",
     "typecheck": "tsc --noEmit",
     "format": "prettier --write \"src/**/*.{ts,js,css,html}\"",
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,6 +16,13 @@ import { log } from "./services/logger.js";
 import type { MemoryType } from "./types/index.js";
 import { getLanguageName } from "./services/language-detector.js";
 import type { MemoryScope } from "./services/client.js";
+import {
+  createV2Client,
+  setConnectedProviders,
+  setHostFetch,
+  setV2Client,
+} from "./services/ai/opencode-provider.js";
+import { getHostClientConfig } from "./services/ai/opencode-host-config.js";
 
 function logAutoCaptureProviderStatus(): void {
   if (!CONFIG.autoCaptureEnabled || CONFIG.autoCaptureProviderStatus.ready) return;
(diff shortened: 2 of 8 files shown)


GraDea:
@tickernelz @x-Spartacus, could you review the changes carefully

tickernelz:
Merged into `main` as part of the v2.19.0 batch release. Verified on the fully-merged tree: `bun run typecheck`, `bun run build`, and `bun test` (197 pass / 0 fail) all green. Thanks!

Issue #170, opened 2026-07-14
"User profile updated" infinitely


LuisPeregrina:
I got this message to appear at least 4 times in 5 minutes. Nothing was done, no conversation, nothing.

<img width="895" height="171" alt="Image" src="https://github.com/user-attachments/assets/dd063c78-e1f4-44bf-9bc9-e2e90edb2ff3" />

They all happened in a posibility of 30 minutes, first one on minute 1, then 30 seconds later, another 30 seconds, then a minute
They always said 10 prompts, even though it is a new session with one prompt.

Since this makes use of the model used in opencode, I think this comsumes tokens uncessesarily. These should work on a number of prompts basis, not every 30 seconds.

Not sure if this is considered a bug or feature, or if these correspond to autoCaptureIterationTimeout and autoCaptureMaxIterations, maybe we can update the readme if they are.

sjclayton:
Yeah, this is totally unusable as it is... it just keeps automatically sending everything over and over again, until the context it's sending is so large it overflows the models context window entirely, at which point the model just starts constantly erroring out because the input is too large. 

Maybe this problem is unrelated to this issue exactly... but it may be partially related in some other way.

cristeaadrian:
I think I tracked the cause of your issue here: https://github.com/tickernelz/opencode-mem/issues/176, but I'm not sure if we're hitting the same problem or just similar ones.

EyJunge1:
This should be fixed now via #191 — the infinite “User Profile Updated” loop is stopped by skipping internal capture sessions and not re-ingesting the profile-analysis prompt echo.

Could you check after updating whether a normal idle/profile cycle looks good on your side? If everything works, it would be great if a maintainer could close this. Thanks for reporting!

PR #171, opened 2026-07-15
Prevent duplicate prompt rows on repeated delivery

lovewave02:
## What changed

`chat.message` can deliver the same session message more than once. Previously,
each `savePrompt()` call created a new row, so one logical prompt could be
counted multiple times by profile learning.

This makes prompt persistence idempotent for each `(session_id, message_id)`
pair and returns the existing prompt id on repeated delivery. It preserves the
row's existing capture state and still treats the same message id in different
sessions as separate prompts.

This does not remove historical duplicate rows or change profile-learning
timing; it prevents new duplicates from this delivery path.

## Verification

- focused regression suite: 11 passed
- full test suite: 202 passed
- `bun run typecheck`
- `bun run format:check`
- `bun run build`

Addresses #170


tickernelz:

PR #175, opened 2026-07-17
fix: reuse authenticated OpenCode client transport

### LINKED ISSUE #174: [Bug] opencodeProvider fails with password-protected OpenCode server
### What's wrong? Maybe some steps to reproduce.

The `opencodeProvider` auto-capture path fails with an unauthorized error when the OpenCode server is started with a password. This is how I run OpenCode through OpenChamber.

#### Steps to reproduce

1. Start the OpenCode server with password authentication enabled through OpenChamber.
2. Install and enable `opencode-mem`.
3. Configure the OpenCode-backed provider, for example:

   ```jsonc
   {
     "opencodeProvider": "provider-name",
     "opencodeModel": "model-name"
   }
   ```

4. Use OpenCode normally and wait for the idle auto-capture operation.

#### Actual behavior

The auto-capture request fails with `unauthorized` when it calls the OpenCode session API. The OpenCode-backed memory capture therefore does not work with a password-protected server.

#### Expected behavior

The plugin should reuse the authenticated OpenCode transport, or provide a documented way to pass the server authentication to its session API requests, so `opencodeProvider` works when the server requires a password.

### Where does it happen?

OpenCode-backed AI provider requests used by auto-capture, specifically the temporary session create/prompt/delete requests.

The current source appears to:

- derive the server URL and an optional host fetch in `src/index.ts`;
- create a separate SDK client with only `{ baseUrl }` in `src/services/ai/opencode-sdk-client.ts`;
- send the raw session requests in `src/services/ai/opencode-provider.ts` with only `Content-Type` headers.

This appears to omit or lose the password-based server authentication when the host fetch is unavailable or does not forward it.

### Version (if known)

- opencode-mem: Not known
- OpenCode: Not known
- OpenChamber: Not known

### Screenshots / recordings (optional)

N/A

### Logs (optional)

```text
unauthorized
```

No credentials or authorization headers are included here.

catan271:
## Summary

- Reuse the OpenCode host SDK transport, including custom fetch and headers, when creating the v2 client.
- Prefer the authenticated SDK session methods for structured output when available.
- Retain the raw HTTP fallback for older SDK builds.
- Add regression coverage for host headers and response content types.

## Validation

- `bun test tests/opencode-provider.test.ts tests/plugin-host-config.test.ts`
- `bun run typecheck`
- `bunx prettier --check src/index.ts src/services/ai/opencode-host-config.ts src/services/ai/opencode-provider.ts src/services/ai/opencode-sdk-client.ts tests/opencode-provider.test.ts tests/plugin-host-config.test.ts`

Closes #174

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -46,7 +46,12 @@ export async function configureOpencodeHostTransport(ctx: {
 
   const serverUrl = hostConfig.baseUrl ?? ctx.serverUrl;
   if (serverUrl) {
-    setV2Client(createV2Client(serverUrl));
+    setV2Client(
+      createV2Client(serverUrl, {
+        fetch: hostConfig.fetch,
+        headers: hostConfig.headers,
+      })
+    );
   }
 }
 
--- a/src/services/ai/opencode-host-config.ts
+++ b/src/services/ai/opencode-host-config.ts
@@ -1,6 +1,7 @@
 export type HostClientConfig = {
   readonly baseUrl: string | undefined;
   readonly fetch: typeof fetch | undefined;
+  readonly headers?: RequestInit["headers"];
   readonly clientKeys: readonly string[];
   readonly sdkConfigCount: number;
 };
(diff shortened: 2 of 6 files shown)


tickernelz:

PR #178, opened 2026-07-18
refactor(memory): use XML markers for injected context

GCS-ZHN:
## Summary
Replace the plain-text `[MEMORY]` header in injected context with explicit XML tags so the model can clearly distinguish reference material from the user's message.

## Changes
- `src/services/context.ts` — wrap the injected block in `<memory_context>...</memory_context>` with a header reminding the model the contents are background context. Nest `<user_profile>` and `<project_knowledge>` sections; each memory item becomes `<memory relevance="N%">...</memory>`.
- `src/services/user-profile/profile-context.ts` — emit profile categories as `<user_preferences>`, `<user_patterns>`, `<user_workflows>`, and `<learning_paths>` with their items as nested elements carrying `category` / `topic` / `frequency` attributes. Add `escapeXmlText` / `escapeXmlAttr` helpers and apply them to all user-supplied strings so profile content cannot break out of the tags.

## Behavior preserved
- `formatContextForPrompt` still returns `""` when there is nothing to inject, so the `if (memoryContext)` guard in `src/index.ts:279` behaves identically.
- `getUserProfileContext` still returns `string | null`; `null` is returned when no profile or no injectable items exist.
- The `parts` array name, `forEach` + `parts.push` calling style, and the workflow steps formatting expression are kept as-is. Only the pushed string contents change.

## Motivation
Models that consume injected context benefit from explicit structural markers: tags make it trivial to identify the start/end of the injected region, attributes carry similarity / category / frequency without parsing prose, and escaping prevents profile content (some of which is written by AI cleanup passes) from breaking out of the wrapper.

## Test plan
- No automated tests target `formatContextForPrompt` output today. Suggested follow-up: a unit test asserting the XML structure (see PR comment).
- Manual: trigger chat.message with a user that has a profile and project memories; confirm the injected part begins with `<memory_context>` and closes with `</memory_context>`.

### DIFF
--- a/src/services/context.ts
+++ b/src/services/context.ts
@@ -15,28 +15,33 @@ export function formatContextForPrompt(
   userId: string | null,
   projectMemories: MemoriesResponseMinimal
 ): string {
-  const parts: string[] = ["[MEMORY]"];
+  const parts: string[] = [];
 
   if (CONFIG.injectProfile && userId) {
     const profileContext = getUserProfileContext(userId);
     if (profileContext) {
-      parts.push("\n" + profileContext);
+      parts.push(`<user_profile>\n${profileContext}\n</user_profile>`);
     }
   }
 
   const projectResults = projectMemories.results || [];
   if (projectResults.length > 0) {
-    parts.push("\nProject Knowledge:");
+    parts.push("<project_knowledge>");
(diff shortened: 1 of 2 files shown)


tickernelz:

PR #180, opened 2026-07-20
Fix path traversal, missing API auth, and stored XSS in local web server/UI

RasAlGhul96:
## Summary

Security audit of the local HTTP API and bundled web UI. Full writeup in `SECURITY_AUDIT.md` (added by this PR). Three exploitable issues fixed with regression tests, two lower-severity issues documented as recommendations only.

- **Path traversal (critical):** `containerTag` from `POST /api/memories` reached shard-path construction (`path.join`) with no character validation, allowing directory creation and SQLite file writes outside `~/.opencode-mem/data`. `extractScopeFromTag()` and the duplicate parsing in `migration-service.ts` now reject anything but the alphanumeric sha256-hex hash the plugin itself always generates.
- **No authentication on the local API (high):** CORS (which lets requests without an `Origin` header through) was the only gate on `/api/*`, so any non-browser local client, or a CSRF-style request from a malicious web page, had full read/write/delete access to memories and the user profile. Added a random per-install token (`src/services/auth-token.ts`) required on every `/api/*` request, injected into the served `index.html` so the bundled UI keeps working transparently.
- **Stored XSS (high):** `renderUserProfile()` injected `profile.displayName` into `innerHTML` without `escapeHtml()`, unlike every other user-controlled field in `app.js`. `displayName` can come from an untrusted project's `userNameOverride` config, which loads automatically when that project is opened.

Not fixed (documented only): unpinned/non-SRI CDN scripts in `index.html`, and the Gemini provider sending its API key via query string (matches Google's own API design).

## Test plan

- [x] `bun install && bunx tsc --noEmit` — no type errors
- [x] `bun test` — all pre-existing passing tests remain green; new regression tests (`tests/api-handlers-container-tag-traversal.test.ts`, `tests/web-userprofile-xss.test.ts`) pass against the patch and were confirmed to fail against the pre-patch code
- [x] Confirmed remaining test failures in the sandbox (EPERM on log file, `dist/plugin.js`-dependent tests) are pre-existing and unrelated, by reproducing them on an unpatched checkout

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/SECURITY_AUDIT.md
+++ b/SECURITY_AUDIT.md
@@ -1,0 +1,56 @@
+# Security Audit — opencode-mem
+
+**Scope:** `tickernelz/opencode-mem` @ commit `0998c69` (main). Local HTTP API, CORS policy, secret handling, SQLite layer, web UI, config/JSONC parsing, migration/cleanup services, dependencies. Read-only audit followed by targeted fixes for the three highest-severity findings.
+
+## Findings
+
+### 1. CRITICAL — Path traversal via unauthenticated `containerTag` (fixed)
+
+`extractScopeFromTag()` in `src/services/api-handlers.ts` split the client-supplied `containerTag` on `_` and used everything after the second `_` as a `hash`, with no character validation. That hash was concatenated into a shard filename in `shard-manager.ts:getShardPath()` and passed through `path.join()`, then `connection-manager.ts:getConnection()` created any missing parent directories and opened/created a SQLite database at the resulting path.
+
+A request such as `POST /api/memories {"content":"x","containerTag":"project_x_../../../../../../Temp/pwn"}` creates directories and a SQLite file outside `~/.opencode-mem/data`. The same unsanitized parsing was duplicated in `migration-service.ts`'s re-embed path.
+
+**Fix:** `extractScopeFromTag()` now validates the hash segment against `^[a-zA-Z0-9]+$` (matching the sha256-hex format the plugin itself always generates) and throws on anything else; the duplicate parsing in `migration-service.ts` got the same guard. Regression test: `tests/api-handlers-container-tag-traversal.test.ts`.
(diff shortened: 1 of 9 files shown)


tickernelz:
I tried to merge this PR after integrating the newer PRs in this batch, but it now has a merge conflict in:

- `src/services/web-server.ts`

Please rebase or merge the latest `main`, resolve the conflict, and make sure the security changes remain covered by tests. We can merge it after that.

RasAlGhul96:
Rebased on latest `main` and resolved the conflict in `src/services/web-server.ts`.

The conflict was between this PR's mandatory per-install auth token gate on `/api/*` and the `WebAuth` (optional HTTP Basic Auth) + `/api/health` endpoint added on `main` in the meantime. Resolution keeps both mechanisms layered together:

- CORS origin check → OPTIONS preflight → optional `WebAuth` Basic Auth check → mandatory auth-token check on all `/api/*` routes.
- `/api/health` is exempted from the token check (same as it already is from `WebAuth`), since it carries no memory data and the bundled UI's `checkAuthWarning()` probes it unauthenticated to decide whether to show the "no auth" banner.
- `checkServerAvailable()` now hits `/api/health` (per `main`'s change) without needing to attach the auth token, since that route is unauthenticated by design.

Verified: `bunx tsc --noEmit` clean, and `bun test` shows the same pass/fail counts as an unpatched checkout of `main` (184 pass / 32 fail — the failures are pre-existing sandbox-only ENOENT log issues unrelated to this change, confirmed by running the same failing files against unpatched `main`). All security-relevant tests (`web-server-cors`, `web-auth`, `api-handlers-container-tag-traversal`, `web-userprofile-xss`, `tool-scope`) pass: 35/35.

tickernelz:

PR #181, opened 2026-07-21
feat(web): HTTP Basic Auth for the web UI

GCS-ZHN:
When the web UI is bound to a non-loopback address (e.g. `0.0.0.0` so it is reachable from the LAN), any browser on the same network can otherwise read and edit memories directly. This PR gates the UI on HTTP Basic Auth so the obvious threat model is closed off by default.

**What changes:**

- New config fields `webServerAuthPassword` / `webServerAuthUsername` turn the gate on. Username defaults to the current OS user; password accepts the same `env://` and `file://` shorthands as `memoryApiKey`.
- `src/services/web-auth.ts` parses `Authorization: Basic …`, compares credentials with `timingSafeEqual`, and serves a 401 + `WWW-Authenticate` challenge with `Cache-Control: no-store`.
- `src/services/cors.ts` now accepts non-loopback browser origins when auth is on, so tools hosted on other origins can talk to the API once they authenticate. `/api/health` is unauthenticated and is reused as the health probe by `checkServerAvailable`.
- The frontend polls `/api/health` on load; if `authEnabled` is false and the page is served from a non-loopback hostname it shows a short banner pointing at `webServerAuthPassword`.
- Tests cover the credential check, the challenge format, and the new CORS option.

**Notes:**

- The branch is rebased onto `upstream/main` so the diff is just the auth feature (no memory-xml-markers mix-in).
- Browser behaviour for the threat model: the browser caches Basic Auth credentials in memory, and closing all browser windows discards them. Reopening the browser requires signing in again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,31 @@ The plugin creates a full commented template at this path on first startup. This
 - `scope: "all-projects"`: query `search` / `list` across all project shards.
 - `memory.defaultScope` sets the default query scope when no explicit scope is provided.
 
+### Web UI HTTP Basic Auth
+
+When `webServerHost` is set to anything other than loopback (for example `0.0.0.0`), the web UI is reachable by anyone on the network. To keep your memories off the LAN, gate the web server with HTTP Basic Auth via the same config file used for everything else:
+
+```jsonc
+{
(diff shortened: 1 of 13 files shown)


tickernelz:

PR #186, opened 2026-07-22
fix(user-profile-analysis): cap context sent to profile LLM

### LINKED ISSUE #176: Profile learning recursively re-ingests its own analysis prompt

4nibhal:
Fixes #176 (partial)

## What this addresses

Item #3 of the suggested fixes in #176: cap the total assembled prompt size that `buildUserAnalysisContext` ships to the profile-analysis LLM. Today the function concatenates the template + all N user prompts with no size guard, so a single oversized prompt or a large backlog can push the payload past the model window.

This PR adds the total-size cap (default 32 KB). It does **not** add the per-item cap or skip-analysis-prompt logic from items #1 and #2 — those are a separate, deeper surgery on the `prompts.map((p, i) => \`${i + 1}. ${p.content}\`).join("\n\n")` line and will come as a follow-up if there is interest.

## Changes
- `src/services/user-memory-learning.ts` (+9/-2): introduce a local `truncate(s)` helper and apply it to both branches (`base` alone, and `base + validationPrompt`).
- `src/config.ts` (+4): new optional knob `userProfileMaxContextBytes` wired through `OpenCodeMemConfig`, `DEFAULTS`, and `buildConfig` in the same shape as `userProfileAnalysisInterval`.
- `README.md` (+1): added the new field to the Configuration Essentials example block.

## Verification
- `bun run typecheck` — clean
- `bun run build` — clean

## Manual check
Set `userProfileMaxContextBytes: 1024` in `~/.config/opencode/opencode-mem.jsonc` and trigger a profile analysis with enough prompts to exceed 1 KB. The LLM payload should be truncated and include the marker `[... context truncated to userProfileMaxContextBytes ...]`.

## Background

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ The plugin creates a full commented template at this path on first startup. This
   "showErrorToasts": true,
 
   "userProfileAnalysisInterval": 10,
+  "userProfileMaxContextBytes": 32768,
   "maxMemories": 10,
 
   "compaction": {
--- a/src/config.ts
+++ b/src/config.ts
@@ -60,6 +60,7 @@ interface OpenCodeMemConfig {
   deduplicationEnabled?: boolean;
   deduplicationSimilarityThreshold?: number;
   userProfileAnalysisInterval?: number;
+  userProfileMaxContextBytes?: number;
   userProfileDisplayPreferences?: number;
   userProfileDisplayPatterns?: number;
   userProfileDisplayWorkflows?: number;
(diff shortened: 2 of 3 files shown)


tickernelz:

PR #190, opened 2026-07-25
fix(ai-cleanup): make opencodeProvider path succeed

### LINKED ISSUE #177: AI Cleanup never succeeds with opencodeProvider
## Summary

AI Cleanup (`POST /api/user-profile/ai-cleanup`) never succeeds with `opencodeProvider`/`opencodeModel` (no external `memoryApiUrl`). Every failure surfaces as `Error: No AI provider configured for profile cleanup` even though the provider is configured and working (auto-capture and profile learning use it fine), because `callAICleanup` catches any error from the opencode path and falls through to the unset `memoryModel`/`memoryApiUrl` branch.

Five stacked bugs, each hidden behind the next. Four are in `callViaOpencodeWithClient` (`ai-cleanup.js`), one is in the web UI (`web/app.js`):

1. **`noReply: true`** on the `session.prompt` call suppresses assistant generation, so no response is produced. The working auto-capture path avoids `noReply` for this reason. Set it to `false`.
2. **Reads a nonexistent field.** The prompt response is `{ info: AssistantMessage, parts: Array<Part> }`, and `AssistantMessage` has no `text` field (per SDK types). Reading `info.text` yields `undefined`, then `"AI response did not contain valid JSON"`. Read the text from `promptResult.data.parts` filtered by `type === "text"`, as the rest of the codebase does.
3. **Server timeout too short.** The `session.prompt` race uses `TIMEOUT_MS = 120000`, but a real cleanup on a sizable profile takes longer, so it aborts. Raise it.
4. **Client default timeout too short.** `fetchAPI` in `web/app.js` defaults the `/ai-cleanup` fetch to 180000ms.
5. **Explicit client timeout overrides the default.** The cleanup call passes an explicit `timeout: 180000`, which wins over the `fetchAPI` default via `options.timeout || ...`. This is the one that actually keeps hanging the modal: the browser aborts before the server responds, so the loading spinner never clears even though the server completed. Raise this explicit value too.

## Repro

1. Configure `opencodeProvider` + `opencodeModel`, no external memory API.
2. Confirm auto-capture / profile learning work.
3. Web UI, click "AI Cleanup".
4. Fails or hangs; log chain: `prompt timeout`, then `AI response did not contain valid JSON`, then `No AI provider configured`.

Verified: with all five fixes applied, cleanup completes and the diff renders for review/apply.

## Environment

- opencode-mem v2.19.4, OpenCode v1.18.3, macOS
- `opencodeProvider: amazon-bedrock`, `opencodeModel: global.anthropic.claude-sonnet-5`, no external memory API

EyJunge1:
## Summary
- Fixes AI Cleanup failing with `opencodeProvider` by stopping reply suppression, reading JSON from text parts, raising cleanup timeouts to 5m, and surfacing OpenCode errors instead of masking them as a missing provider.
- Closes #177

## Test plan
- [x] `bun test tests/ai-cleanup.test.ts` (8/8 pass)
- [x] `bun run typecheck` + `bun run build` + `bun run format:check`
- [x] Full suite after build: `238 pass / 0 fail`
- [x] **Live:** Triggered `POST /api/user-profile/ai-cleanup` against local OpenCode (`opencodeProvider=opencode`, `opencodeModel=deepseek-v4-flash-free`) — `success: true` in ~14s; log shows `session.create` + `session.prompt done` via OpenCode; result merged 5 duplicate groups (`kept:1`, `merged:5`, `removed:0`)
- [x] **Live:** Web UI stayed healthy (`/api/health` 200 before and after cleanup); `/ai-cleanup` path uses 300000ms timeout

### DIFF
--- a/src/services/user-profile/ai-cleanup.ts
+++ b/src/services/user-profile/ai-cleanup.ts
@@ -189,20 +189,16 @@ function formatForAI(item: IndexedProfileItem): Record<string, unknown> {
 async function callAICleanup(
   prompt: string
 ): Promise<{ profile: IndexedProfile; mapping: AIMapping }> {
-  // Use opencode internal session when opencodeProvider is configured (same pattern as auto-capture)
+  // Use opencode internal session when opencodeProvider is configured (same pattern as auto-capture).
+  // When the client is available, surface OpenCode errors instead of masking them as
+  // "No AI provider configured" via a silent fallback (#177).
   if (CONFIG.opencodeProvider && CONFIG.opencodeModel) {
-    try {
-      const { getV2Client } = await loadOpencodeProvider();
-      const v2Client = getV2Client();
-      if (v2Client) {
-        const result = await callViaOpencodeWithClient(v2Client, prompt);
-        if (result) return result;
-      }
-    } catch (e) {
-      log("AI cleanup: opencode session failed, falling back to external API", {
-        error: String(e),
-      });
+    const { getV2Client } = await loadOpencodeProvider();
+    const v2Client = getV2Client();
+    if (v2Client) {
+      return callViaOpencodeWithClient(v2Client, prompt);
     }
+    log("AI cleanup: opencode client unavailable, falling back to external API");
   }
 
   if (CONFIG.memoryModel && CONFIG.memoryApiUrl) {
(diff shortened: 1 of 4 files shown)


tickernelz:

PR #191, opened 2026-07-25
fix(profile-learning): stop infinite User Profile Updated loop

### LINKED ISSUE #170: "User profile updated" infinitely

EyJunge1:
## Summary
- Stops the infinite "User Profile Updated" loop by not persisting echoed profile-analysis prompts, ignoring idle events from internal capture sessions, and marking batches after update conflicts.
- Makes idle skip work against the OpenCode host client (path-based `session.get`) and tracks internal capture session IDs so post-delete idle events still match.
- Closes #170

## Test plan
- [x] `bun test tests/internal-capture-sessions.test.ts tests/plugin-host-config.test.ts` (10/10 pass)
- [x] `bun run typecheck` + `bun run build` + `bun run format:check` (prettier fix pushed)
- [x] Full suite after build: `235 pass / 0 fail`
- [x] **Live:** Against OpenCode host — `internal capture session check via session.get` works; repeated `Skipping idle processing for internal capture session` for structured-output sessions (loop guard). Profile learning runs from the real user session idle path (`user-profile-learning: check` / `analyze called`) without scheduling idle→LLM→idle on capture sessions. No repeating `User Profile Updated` toast storm observed across two idle windows.
- [x] **Live:** Ordinary user prompts still captured (`I prefer short TypeScript…`, `Also use snake_case…` persisted in `user_prompts`); normal session title not matched as internal capture (`matched:false`)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -20,8 +20,13 @@ import type { MemoryScope } from "./services/client.js";
 import { getHostClientConfig } from "./services/ai/opencode-host-config.js";
 import { loadOpencodeProvider } from "./services/ai/opencode-provider-loader.js";
 
-/** Title used for transient structured-output sessions (capture / profile learning). */
-export const INTERNAL_CAPTURE_SESSION_TITLE = "opencode-mem capture";
+import {
+  INTERNAL_CAPTURE_SESSION_TITLE,
+  isInternalCaptureSessionTitle,
+  isTrackedInternalCaptureSession,
+} from "./services/ai/internal-capture-sessions.js";
+
+export { INTERNAL_CAPTURE_SESSION_TITLE, isInternalCaptureSessionTitle };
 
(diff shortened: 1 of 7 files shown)


tickernelz:

PR #195, opened 2026-07-27
fix(opencode): resolve opencodeModel "inherit" on all structured-output paths

anagnorisis2peripeteia:
## Summary

`opencodeModel: "inherit"` (from #151) only expanded to a concrete model on the **auto-capture** path. Profile learning, conflict/dedup checks, description evolution, and other `generateStructuredOutput` callers still sent `modelID: "inherit"` to OpenCode, which fails with:

```text
ProviderModelNotFoundError: Model not found: local/inherit
```

(and the same shape for any configured provider).

This expands `inherit` centrally inside `generateStructuredOutput`:

1. Prefer a prompt-recorded model when provided (auto-capture / `chat.params`)
2. Else fall back to the most recent entry in OpenCode's `model.json` for the configured provider
3. Else throw a clear configuration error instead of a provider lookup failure

## Test plan

- [x] `bun test tests/opencode-provider.test.ts` (26 pass, including 5 new inherit cases)
- [x] With `"opencodeModel": "inherit"` configured, trigger profile learning / auto-capture and confirm no `Model not found: <provider>/inherit` toasts
- [x] Confirm capture still prefers the per-prompt recorded model when `chat.params` has run

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ Supported providers: any provider listed by `opencode providers list` (e.g. `ant
 
 If `opencodeProvider` and `opencodeModel` are set, they take precedence over the manual `memoryProvider` settings below.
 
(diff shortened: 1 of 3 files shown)


tickernelz:

PR #198, opened 2026-07-28
fix(embedding): pin onnxruntime-node for Intel Mac nested installs

### LINKED ISSUE #184: [Bug] Intel Mac still broken on v2.20.0: onnxruntime override is ignored in OpenCode's nested install
## Summary

Related to #158 (fixed by #163). **This is not a duplicate**: #163's intended fix (pin `onnxruntime-node` to `1.22.0` via `overrides`) does not take effect under OpenCode's actual plugin dependency installation model, so **Intel Mac (darwin/x64) local embedding remains broken on opencode-mem v2.20.0**.

## Environment

| Item | Value |
|------|--------|
| OS / arch | macOS Intel (`darwin/x64`) |
| OpenCode | 1.18.3 |
| opencode-mem | **2.20.0** |
| Embedding | default local: `Xenova/nomic-embed-text-v1` |

## Actual behavior

Clean/direct OpenCode install under:

```text
~/.cache/opencode/packages/opencode-mem@2.20.0
```

resolves `@huggingface/transformers@4.2.0` → nested `onnxruntime-node@1.24.3`, which ships **only `darwin/arm64`** and **lacks `darwin/x64`**.

Exact error:

```text
Cannot find module '../bin/napi-v6/darwin/x64/onnxruntime_binding.node'
```

Followed by:

```text
Failed to initialize embedding model
Plugin warmup failed
```

Memory feature stays stuck in initializing.

## Expected behavior

PR #163's pin to `onnxruntime-node@1.22.0` should make Intel local embedding work (native binding present for `darwin/x64`).

## Root cause evidence

The published `opencode-mem@2.20.0` package places:

```json
"overrides": {
  "onnxruntime-node": "1.22.0"
}
```

in the **nested dependency package** (opencode-mem itself), not as something the **install root** applies.

npm/Arborist **only honors root-level `overrides`**. OpenCode's plugin loader installs the plugin tree such that this nested override is **ignored**, so resolution stays on `onnxruntime-node@1.24.3` (no x64 binary). OpenCode's cache also **reuses** the already-installed tree, so a package-only override change does not self-heal existing installs.

## Verified workaround

Replacing **both**:

1. the top-level resolved `onnxruntime-node`, and  
2. `@huggingface/transformers/node_modules/onnxruntime-node`

with **`1.22.0`** provides the `x86_64` binding + dylib, makes `require` succeed, and yields:

```text
Embedding model warmed up
```

with no subsequent warmup errors.

## Suggested fixes

1. Make `onnxruntime-node@1.22.0` an **effective direct/bundled** dependency (or otherwise force resolution at the install root), not a nested-only `overrides` entry.  
2. Add a **post-resolve platform artifact check** (fail fast with a clear message if `darwin/x64` binding is missing).  
3. Until fixed, **document remote embedding fallback** for Intel Mac users.

Upstream context (missing platform binary packaging): [microsoft/onnxruntime#27961](https://github.com/microsoft/onnxruntime/issues/27961)

## Reproduction (concise)

1. On macOS Intel, use OpenCode 1.18.3 with opencode-mem **2.20.0** and default local embedding.  
2. Let OpenCode install the plugin to `~/.cache/opencode/packages/opencode-mem@2.20.0`.  
3. Inspect nested resolution: `@huggingface/transformers@4.2.0` → `onnxruntime-node@1.24.3` (arm64-only layout; no `darwin/x64`).  
4. Trigger memory/embedding warmup → observe the exact `napi-v6/darwin/x64/onnxruntime_binding.node` module-not-found error and warmup failure above.  
5. (Optional) Manually install `onnxruntime-node@1.22.0` into both resolution locations described under Workaround → warmup succeeds.

Happy to provide additional sanitized logs if needed.

EyJunge1:
## Summary

- Fixes #184 by making `onnxruntime-node@1.22.0` a direct dependency so OpenCode nested plugin installs still get a `darwin/x64` binding.
- Adds a resolve shim before loading transformers, plus a fail-fast embedding init error path with a clear remote-embedding hint.
- Keeps the existing `overrides` pin and documents the nested-install reason for the direct dependency.

## Test plan

- [x] `bun test tests/onnxruntime-resolve.test.ts tests/package-dependencies.test.ts` (7 pass)
- [ ] On Intel Mac, clear `~/.cache/opencode/packages/opencode-mem@*` and reinstall the plugin (needs real `darwin/x64` hardware; not verified here)
- [ ] Confirm local embedding warmup succeeds with `onnxruntime-node@1.22.0` (needs real `darwin/x64` hardware; not verified here)
- [x] Confirm a missing x64 binding fails fast with a remote-embedding hint instead of leaving memory stuck initializing (covered by `intel mac missing-binding message mentions remote embedding fallback` + `getEmbeddingInitError()` path)

### DIFF
--- a/.github/workflows/embedding-backend.yml
+++ b/.github/workflows/embedding-backend.yml
@@ -13,6 +13,7 @@ on:
       - "package.json"
       - "bun.lock"
       - "src/services/embedding.ts"
+      - "src/services/onnxruntime-resolve.ts"
       - "scripts/verify-embedding-backend.mjs"
       - ".github/workflows/embedding-backend.yml"
   workflow_dispatch:
@@ -23,7 +24,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-15, windows-latest]
+        # macos-13 = Intel (darwin/x64); macos-15 = Apple Silicon (darwin/arm64).
+        # Intel must keep resolving onnxruntime-node@1.22.0 (#184).
+        os: [ubuntu-latest, macos-13, macos-15, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v5
--- a/.github/workflows/embedding-backend.yml
+++ b/.github/workflows/embedding-backend.yml
@@ -24,9 +24,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # macos-13 = Intel (darwin/x64); macos-15 = Apple Silicon (darwin/arm64).
+        # macos-15-intel = Intel (darwin/x64); macos-15 = Apple Silicon (darwin/arm64).
         # Intel must keep resolving onnxruntime-node@1.22.0 (#184).
-        os: [ubuntu-latest, macos-13, macos-15, windows-latest]
+        os: [ubuntu-latest, macos-15-intel, macos-15, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v5
--- a/bun.lock
+++ b/bun.lock
@@ -11,6 +11,7 @@
         "@opencode-ai/sdk": "^1.18.4",
         "franc-min": "^6.2.0",
         "iso-639-3": "^3.0.1",
+        "onnxruntime-node": "1.22.0",
         "zod": "^4.4.3",
       },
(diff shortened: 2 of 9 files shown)


tickernelz:

- `bun run typecheck`
- `git diff --check`

tickernelz:

Issue #98, opened 2026-04-18
 feat: Add optional task-specific prefixes for nomic-embed-text-v1.5


ZeR020:
The nomic-embed-text-v1.5 model performs better with task-specific prefixes:
- `search_document:` for storing memories
- `search_query:` for searching
Currently opencode-mem sends raw text, which works but isn't optimal for Nomic models.
Request: Add an opt-in config option `embeddingUseTaskPrefixes` that applies these prefixes when enabled.
Docs: https://huggingface.co/nomic-ai/nomic-embed-text-v1.5#task-instruction-prefixes

EyJunge1:
Fixed by #199 — opt-in `embeddingUseTaskPrefixes` applies Nomic-style `search_document:` / `search_query:` prefixes when enabled.

Closing manually because the PR used “Implements #98” rather than “Fixes/Closes #98”, so GitHub did not auto-close on merge.

PR #199, opened 2026-07-28
feat(embedding): add opt-in Nomic task prefixes

EyJunge1:
## Summary

- Implements #98 with an opt-in `embeddingUseTaskPrefixes` config flag (default `false`).
- Applies Nomic-style `search_document:` / `search_query:` prefixes at store and query call sites when enabled.
- Documents that existing memories should be re-indexed after enabling so store and query vectors stay aligned.

## Test plan

- [x] `bun test tests/embedding-task-prefixes.test.ts tests/config.test.ts` (32 pass)
- [x] Confirm default behavior is unchanged with `embeddingUseTaskPrefixes` unset/false (default `false` + prefix helper leaves text unchanged when disabled)
- [x] Confirm document/query prefixes are applied when enabled (`search_document:` / `search_query:` unit coverage; call sites pass `{ task: "document" | "query" }`)
- [x] Confirm README/config guidance mentions the re-index requirement after enabling

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,9 @@ The plugin creates a full commented template at this path on first startup. This
   "userEmailOverride": "user@example.com",
   "userNameOverride": "John Doe",
   "embeddingModel": "Xenova/nomic-embed-text-v1",
+  // Optional Nomic task prefixes (search_document: / search_query:). After enabling,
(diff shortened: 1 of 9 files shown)


tickernelz:

- `bun run typecheck`
- `git diff --check`

tickernelz:

PR #200, opened 2026-07-28
ci: build artifacts before package smoke tests

tickernelz:
## Root causes

1. `Platform Package Smoke` ran `bun test` before `bun run build`; clean runners had no `dist/` for contract/bundle tests.
2. Stable `libsql` 0.5.x can retain prepared-statement/native file handles after `Database.close()` on Windows, causing `EBUSY`/`EPERM` during migration swaps and test cleanup (upstream: https://github.com/tursodatabase/libsql-js/issues/228).
3. `closeConnection()` did not wait for an in-flight open in the manager's `pending` map, allowing the open to finish after “close” and re-lock the database during a swap.
4. `macos-13` was retired by GitHub Actions in December 2025 and no longer starts jobs.
5. npm ignores `overrides` declared by an installed dependency. The packed-package smoke therefore installed `onnxruntime-node@1.24.3`, which has no Darwin x64 binary, instead of the package's pinned `1.22.0` runtime (upstream: https://github.com/microsoft/onnxruntime/issues/27961).
6. Windows Git worktree identity could differ between long and 8.3 path spellings.

## Fix

- Build package artifacts before running the test suite.
- Restrict forced handle collection to Windows and add bounded retries for lock-like file mutation failures.
- Apply retries to migration swaps, archive/delete paths, rollback paths, and staged-file cleanup without weakening transaction/error semantics.
- Release the local legacy DB reference before close/GC.
- Serialize per-path close/open operations, wait for pending opens before closing, and cover the race with a deterministic fake-client test that does not add native teardown state.
- Centralize strict Turso test cleanup around the same close/retry behavior.
- Run the libSQL vector smoke in a child process so the parent removes test files only after native handles are released.
- Replace retired `macos-13` with `macos-15-intel` while retaining Apple Silicon coverage on `macos-15`.
- Mirror the package's optional ONNX runtime override into the temporary npm smoke root; native loading remains a hard failure.
- Canonicalize Windows Git paths with `realpathSync.native()`.

## Verification

Final head: `71779bd5c0d40f36eacb116e8496ea9f644f6b43`

Local:

- `bun run typecheck`
- `bun run build`
- `bun test` — **269 pass, 0 fail, 591 assertions**
- libSQL vector smoke — **5/5 pass, 0 leftover files**
- Turso-focused suite — **3 consecutive passes (30/30 each)**
- `git diff --check`

GitHub Actions final run: https://github.com/tickernelz/opencode-mem/actions/runs/30358481646

- `ubuntu-latest / package smoke` — success
- `windows-latest / package smoke` — success
- `macos-15 / package smoke` — success
- `macos-15-intel / package smoke` — success
- GitGuardian Security Checks — success

Runtime and CI fixes were published in [v2.21.1](https://github.com/tickernelz/opencode-mem/releases/tag/v2.21.1).

### DIFF
--- a/.github/workflows/platform-smoke.yml
+++ b/.github/workflows/platform-smoke.yml
@@ -43,11 +43,11 @@ jobs:
       - name: Typecheck source
         run: bun run typecheck
 
-      - name: Run test suite
-        run: bun test
-
       - name: Build package artifacts
         run: bun run build
+
+      - name: Run test suite
+        run: bun test
 
       - name: Pack and install published package shape
         shell: bash
--- a/.github/workflows/platform-smoke.yml
+++ b/.github/workflows/platform-smoke.yml
@@ -64,14 +64,17 @@ jobs:
 
           cd "$smoke_dir"
           npm init -y
+          # npm ignores overrides declared by installed dependencies, so mirror the plugin's
+          # native-runtime pin at the temporary root used by this package smoke test.
           override_version=$(node --input-type=module -e '
             import { readFileSync } from "node:fs";
             import { resolve } from "node:path";
             const pkg = JSON.parse(readFileSync(resolve(process.env.GITHUB_WORKSPACE, "package.json"), "utf8"));
             process.stdout.write(pkg.overrides?.["onnxruntime-node"] ?? "");
           ')
-          test -n "$override_version"
-          npm pkg set "overrides.onnxruntime-node=$override_version"
+          if [ -n "$override_version" ]; then
+            npm pkg set "overrides.onnxruntime-node=$override_version"
+          fi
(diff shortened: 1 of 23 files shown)


tickernelz:
Merged and released in [v2.21.1](https://github.com/tickernelz/opencode-mem/releases/tag/v2.21.1).

Final verification:
- Platform Package Smoke: Ubuntu, Windows, macOS ARM, and macOS Intel all passed
- Release workflow: https://github.com/tickernelz/opencode-mem/actions/runs/30359186601
- npm `latest`: `2.21.1`
- Published tarball: 200 files; required Turso runtime modules and close/retry fixes verified

PR #201, opened 2026-07-28
Replace legacy Memory Explorer with React 19

EyJunge1:
## Summary
- replace the legacy vanilla/Svelte Memory Explorer with a React 19 + Vite + Tailwind/shadcn UI
- preserve the REST API, auth-token injection, routing, i18n, themes, memory workflows, profile management, and migrations
- share validated API types with the frontend and serve built assets consistently in compiled and source/test modes

## Test plan
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] `bun test` — 270 tests pass; local Bun 1.3.14 panics during native teardown after reporting completion
- [x] HTTP smoke test: `/`, `/project-memories`, `/user-profile`, `/api/health`, and the hashed React bundle return 200
- [x] repository scan confirms no Svelte source files, Svelte dependencies, or legacy `src/web` assets remain

Made with [Cursor](https://cursor.com)

PR #204, opened 2026-07-30
ci(release): install web dependencies before build

tickernelz:
## Problem

The `v2.22.0` release run failed during `bun run build` because the release workflow installed only root dependencies. Since #201, the root build invokes the `web` TypeScript/Vite build, which requires `web/node_modules`.

Failed run: https://github.com/tickernelz/opencode-mem/actions/runs/30507159676

## Fix

- install root dependencies with `bun install --frozen-lockfile`
- install `web` dependencies from `web/bun.lock` before typecheck/build

## Verification

Reproduced the exact release install/typecheck/build sequence in a clean detached worktree with neither root nor web `node_modules` present:

- root frozen install: pass
- web frozen install: pass
- `bun run typecheck`: pass
- `bun run build`: pass
- `dist/plugin.js`: present
- `dist/web/index.html`: present

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,12 @@ jobs:
 
       - uses: oven-sh/setup-bun@v2
 
-      - run: bun install
+      - name: Install source dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Install web dependencies
+        working-directory: web
+        run: bun install --frozen-lockfile
 
       - run: bun run typecheck
 


tickernelz:
Confirmed in production release `v2.22.0`: the corrected workflow installed both dependency sets, built the package and React UI, published to npm, and created the GitHub Release successfully.

Run: https://github.com/tickernelz/opencode-mem/actions/runs/30507456715

Issue #169, opened 2026-07-14
Auto Capture Failed


LuisPeregrina:
Related to #162 , from https://github.com/anomalyco/opencode/issues/36793

opencode.json
```json
{
  ...
  "plugin": [
    "opencode-mem",
    ...
  ],
  "agent": {
    "plan": {
      "model": "deepseek/deepseek-v4-pro",
      "thinking": {
        "type": "enabled"
      },
      "reasoning_effort": "max"
    },
    "chatter": {
      "model": "deepseek/deepseek-v4-flash",
      "description": "Zero-tool agent for general knowledge and conversation.",
      "prompt": "You are a helpful assistant.",
      "thinking": {
        "type": "disabled"
      }
    }
  }
...
}
```

I don't know where to look for logs.

I removed the opencode cache via `rm -rf ~/.cache/opencode`, started this new session:

<img width="1141" height="379" alt="Image" src="https://github.com/user-attachments/assets/60520500-16bb-4afa-8eea-62ecaf93f33b" />

LuisPeregrina:
NVM the model set in opencodeModel no longer exists.

PR #230, opened 2026-08-02
feat: support model variant in opencode provider path

Ghilteras:
## Problem

`opencode-mem` sends `POST /session/{id}/message` with a model object containing only `providerID` and `modelID`. The opencode SDK/API also accepts an optional `variant` on the model object (e.g. `thinking-off` for `deepseek-v4-flash`).

For custom provider aliases (like an `opencode-go` alias for deepseek) some model variants require the variant to be sent explicitly — omitting it leads to a 401 / `ProviderModelNotFoundError` on those aliases while the same model with a variant works.

## Change

Add an optional `opencodeVariant` config field (string) that is forwarded verbatim as `model.variant` on every opencode structured-output request:

- `src/config.ts`: new `opencodeVariant?: string` config field, parsed next to `opencodeProvider` / `opencodeModel`, plus sample config docs.
- `src/services/ai/opencode-provider.ts`: `generateStructuredOutput` accepts `variant` in opts and propagates it through both transports (v2 SDK client and raw fetch) via `sessionPromptFields`, which adds `variant` to the model object only when set (no behavior change when unset).
- Callers pass `CONFIG.opencodeVariant`: `auto-capture.ts`, `user-memory-learning.ts` (both profile analysis and learning-paths paths), `user-profile/user-profile-manager.ts` (dedup, conflict, description evolution), and `user-profile/ai-cleanup.ts` (direct `session.prompt` call).
- `README.md`: documents the new `opencodeVariant` field.

## Config

```jsonc
"opencodeProvider": "opencode-go",
"opencodeModel": "deepseek-v4-flash",
"opencodeVariant": "thinking-off"
```

## Tests

- `bun run typecheck` passes.
- `bun test`: 347 pass / 0 fail (added 2 tests in `opencode-provider.test.ts` asserting the variant is included in the model object when set and omitted when not, plus 1 config-resolution test).
- `bun run build`: `tsc` dist build passes; the web build fails on pre-existing environment issues unrelated to this change (missing `web/node_modules` + TS 7 `baseUrl` removal), reproducible on a clean checkout.

## Backward compatibility

Fully optional: when `opencodeVariant` is unset, the model object is byte-identical to before.


Ghilteras:
Motivating issue: #169 (Auto Capture Failed — thinking/tool_choice incompatibility on models with thinking enabled). This PR addresses the root gap that issue exposed: opencode-mem could not express a model variant (e.g. thinking-off) in its structured-output calls, so auto-capture/user-profile failed on providers requiring the variant to be explicit. Adds optional opencodeVariant config forwarded as model.variant; no behavior change when unset.

PR #248, opened 2026-08-09
fix: mark compacted memory injection part as synthetic

phil-lipp:
The compaction restore injected the memory dump as a normal text part, so it was echoed into the live session transcript. Mark the part synthetic so the TUI does not render it.

Follow-up to #239, which preserved the session agent on compaction restore but did not mark the injected part synthetic.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -973,7 +973,14 @@ export const OpenCodeMemPlugin: Plugin = async (ctx: PluginInput) => {
           await ctx.client.session.prompt({
             path: { id: sessionID },
             body: {
-              parts: [{ id: `prt-compaction-${Date.now()}`, type: "text", text: memoryContext }],
+              parts: [
+                {
+                  id: `prt-compaction-${Date.now()}`,
+                  type: "text",
+                  text: memoryContext,
+                  synthetic: true,
+                },
+              ],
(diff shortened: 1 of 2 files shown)


lindixu6-hash (approved):
Verified at 30374c3. The change follows the existing synthetic-part convention used by chat.message injection, and the new regression is failure-sensitive.

Local verification with an isolated HOME:
- frozen root and web installs
- compaction target: 9/9 passed
- full suite after build: 408/408 passed
- typecheck, Prettier check, and production build passed
- mutation check: reverting only the implementation while retaining the test produced the expected 1 failure (synthetic was undefined)

The scope is small and preserves the existing agent/noReply behavior.

lindixu6-hash:
Post-merge validation for exact merge commit `25a419a4860eeb802215809202b0204e2309e2f6`: [Platform Package Smoke run 32279260250](https://github.com/lindixu6-hash/opencode-mem/actions/runs/32279260250) passed all six jobs (Ubuntu, Windows, macOS 15/26 on Apple Silicon, and macOS 15/26 Intel). Each job completed dependency installation, typecheck, production build, the full test suite, package creation/install, and native/runtime smoke checks.

PR #249, opened 2026-08-09
fix: deduplicate embedded Tags footer in compaction restore

phil-lipp:
Auto-capture embeds a "Tags: ..." footer inside the memory body (added in #131); compaction restore appends its own canonical Tags line, duplicating the tags in the restored context. Strip the embedded footer when a canonical line follows.

Regression from #131.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1041,14 +1041,55 @@ function formatSearchResults(query: string, results: any, limit?: number): strin
   });
 }
 
+const EMBEDDED_TAGS_FOOTER_RE = /\n*Tags: ([^\n]*)\s*$/;
+
+function normalizeTagsKey(tags: string[]): string {
+  return tags
+    .map((tag) => tag.trim())
+    .filter((tag) => tag.length > 0)
+    .sort()
+    .join("\0");
+}
+
+function stripMatchingEmbeddedTagsFooter(memory: string, tags: string[]): string {
+  const match = memory.match(EMBEDDED_TAGS_FOOTER_RE);
+  if (!match) {
+    return memory;
+  }
+
+  const footerValue = match[1];
+  if (footerValue === undefined) {
+    return memory;
+  }
+
+  const embeddedTags = footerValue
+    .split(",")
+    .map((tag) => tag.trim())
+    .filter((tag) => tag.length > 0);
+
+  if (normalizeTagsKey(embeddedTags) !== normalizeTagsKey(tags)) {
+    return memory;
+  }
+
+  return memory.replace(EMBEDDED_TAGS_FOOTER_RE, "");
+}
+
 function formatMemoriesForCompaction(memories: any[]): string {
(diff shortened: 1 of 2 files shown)


lindixu6-hash (changes requested):
The duplication fix is valid for auto-captured records, and the submitted tests are failure-sensitive, but the current regex removes any trailing `Tags:` line whenever structured tags exist, even when the footer is user-authored and does not match those tags.

For example, a memory ending in `Tags: user-authored evidence` with structured tags `["canonical"]` is rendered without that original line and only `Tags: canonical` remains. Manual/API/imported memories can store arbitrary content plus structured tags, so this is observable content loss.

Please strip the footer only when its normalized value matches the canonical structured tags (or when another reliable auto-capture provenance condition proves it is the generated footer), and add a regression showing a non-matching trailing `Tags:` line is preserved.

Local verification at 9ad5b8e with an isolated HOME:
- frozen root and web installs
- compaction target: 10/10 passed
- full suite after build: 409/409 passed
- typecheck, Prettier check, and production build passed
- mutation check: reverting only the implementation while retaining the tests produced the expected duplicate-footer failure

So the existing duplication regression is covered; the remaining blocker is preserving unrelated user content.

lindixu6-hash:

tickernelz:
Skipped: this PR conflicts with `main` (`tests/compaction-agent-preservation.test.ts`). Please rebase onto the latest `main` and resolve the conflict, then it can be included in a future batch.

tickernelz:
Skipped: this PR conflicts with `main` (`tests/compaction-agent-preservation.test.ts`). Please rebase onto the latest `main` and resolve the conflict — it can be included in a future batch.

lindixu6-hash (approved):
The follow-up at exact head `ca4b842b891faf70f5515893c2d309c83ea6e638` resolves the data-loss blocker from my previous review.

```mermaid
flowchart LR
  A[Stored memory body] --> B{Trailing Tags footer?}
  B -->|No| C[Preserve body]
  B -->|Yes| D{Normalized footer equals structured tags?}
  D -->|Yes| E[Strip embedded duplicate]
  D -->|No| C
  C --> F[Append canonical structured Tags line]
  E --> F
  style E fill:#c8e6c9,color:#1a5e20
  style C fill:#bbdefb,color:#0d47a1
```

The implementation now strips a trailing `Tags:` footer only when its trimmed, order-independent value matches the structured tags. A non-matching user-authored footer remains intact, while the canonical structured line is still appended. The new regression directly covers the previously demonstrated `user-authored evidence` / `canonical` mismatch.

Independent verification:
- `HOME=$PWD/.test-home bun test tests/compaction-agent-preservation.test.ts` — 12 pass, 0 fail;
- `bun run typecheck` — pass;
- Prettier check on both changed files — pass;
- upstream six-platform package-smoke jobs on this head — all pass.

No remaining blockers. Approved.

PR #252, opened 2026-08-13
fix: correctness & data-integrity fixes across user profiles and turso sharding

phil-lipp:
## Summary

Ten correctness / data-integrity bugs found during a code-quality pass, grouped into five reviewable commits (each builds and typechecks independently). No behavior changes beyond fixing the defects; no public API changes.

`bun run typecheck` clean, `bun test` green (adds one regression test).

### `user-profile-manager.ts`
1. **Cross-user profile contamination via the cold-start buffer.**
   `coldBuffer` was a single instance field on the manager singleton. Items buffered during embedding warm-up for one user drained into whichever user's `mergeItems` ran next, merging User A's preferences into User B's profile. Now keyed per `profileId` (`Map`); each merge drains only its own bucket. Legacy unattributed buffer files are dropped on load. *(covered by the new test)*
2. **`learning_paths` stripped on every write.** `createProfile` / `updateProfile` rebuilt `cleanedData` as only `{preferences, patterns,  workflows}`, silently discarding `learning_paths` set by `buildLearningPaths` — the Learning Paths injection feature was effectively dead. Carried through.

### `user-memory-learning.ts`
3. **Unawaited learning-captured marker → duplicate analysis.** The retry-exhausted branch didn't await `markMultipleAsUserLearningCaptured`, so `finally` cleared `isLearningRunning` mid-write; the next cycle re-fetched and re-analyzed the same prompts (token burn), and the rejection was unhandled.
4. **Fire-and-forget `evolveAndUpdate` races serialization (lost update).** It mutates `item.description`/`centroid` in place but was called `.catch(()=>{})`, racing the `JSON.stringify` in `updateProfile` — the evolved description was included or lost nondeterministically. `applyValidations` is now `async` and awaits the evolve so mutation completes pre-serialization.
5. **No `catch` around stored-JSON parses.** `performUserProfileLearning` had `try/finally` but no `catch`; `JSON.parse` of a corrupt `profileData` row rejected the promise (unhandled at the fire-and-forget call site). Added.

### `user-profile/ai-cleanup.ts`
6. **Hallucinated AI ids abort the whole cleanup.** In the merged-group rebuild, a keeper id present only in the model's cleaned output (absent from `originalById`) threw on `originalItem.frequency`, aborting the entire run. Skip such groups. Also added `normalizeAIMapping` so a malformed `mapping` (`kept`/`merged`/`removed`) degrades to a no-op cleanup instead of throwing on `.map`/`.filter`/`.includes`.

### `user-profile/profile-context.ts`
7. **Unguarded `JSON.parse` in the injection hot path.** One corrupt profile row broke context injection for every request. Wrapped in try/catch; returns `null` on parse failure (mirrors `loadColdBuffer`).

### `turso/shard-manager.ts`
8. **Orphaned shard registry row bricks writes.** `createShard` committed the registry `INSERT` before `initShardDb` ran; if init threw (disk full, permissions) the row persisted pointing at an uninitialized file, so the next
   `getWriteShard` failed `isShardValid` and threw "incompatible or corrupt", blocking all writes to that scope. Now initializes the shard DB first (`initShardDb` is idempotent), then inserts.
9. **`getShardByPath` LIKE treats `_` as a wildcard.** `WHERE db_path LIKE '%' || ?` with a filename full of `_` (`user_<hash>_shard_N.db`) matched any char and could return the wrong shard row. Anchored on the `/` separator with escaped LIKE metacharacters (`ESCAPE '\'`).

### `turso/operation-lock.ts`
10. **Unguarded `unlinkSync` in lock cleanup.** `readLiveLock` called `unlinkSync` outside its try/catch; a race (already removed) or a Windows open handle threw ENOENT/EPERM out of `assertNoTursoMigrationInProgress`, falsely blocking writes. Wrapped it.

## Tests

Adds `tests/user-profile-cold-buffer-isolation.test.ts`, asserting that items buffered for one profile during embedding cold-start never drain into another profile's merge, and that each bucket drains only for its own profile.

## Notes for reviewers

- The commits are split by subsystem and can be cherry-picked independently if  you'd prefer to take a subset.
- `normalizeAIMapping` intentionally coerces to safe empty arrays rather than throwing, so a bad model response results in a no-op cleanup (originals preserved) rather than an aborted run — the more data-preserving choice.

### DIFF
--- a/src/services/turso/operation-lock.ts
+++ b/src/services/turso/operation-lock.ts
@@ -30,7 +30,12 @@ function readLiveLock(path: string): LockState | null {
   } catch {
     // Corrupt locks are stale and removed below.
   }
-  unlinkSync(path);
+  try {
+    unlinkSync(path);
+  } catch {
+    // Already removed (race) or held open (Windows) — either way the lock is
+    // gone or unowned; do not let cleanup failure block writes.
+  }
   return null;
 }
 
--- a/src/services/turso/shard-manager.ts
+++ b/src/services/turso/shard-manager.ts
@@ -180,6 +180,13 @@ export class TursoShardManager {
     const storedPath = join(`${scope}s`, basename(fullPath)).replace(/\\/g, "/");
     const now = Date.now();
 
+    // Initialize the shard file BEFORE inserting the registry row. If init throws
+    // (disk full, permissions), the registry stays free of an orphan row that points
+    // at an uninitialized file — such a row would later fail isShardValid on every
+    // getWriteShard and brick all writes to this scope. initShardDb is idempotent.
+    const shardDb = await tursoConnectionManager.getConnection(fullPath);
+    await this.initShardDb(shardDb);
+
     let result;
     try {
       result = await metadataDb.execute(
@@ -204,9 +211,6 @@ export class TursoShardManager {
       }
       throw error;
     }
-
-    const shardDb = await tursoConnectionManager.getConnection(fullPath);
-    await this.initShardDb(shardDb);
 
     return {
       id: Number(result.lastInsertRowid),
@@ -487,9 +491,14 @@ export class TursoShardManager {
   async getShardByPath(dbPath: string): Promise<ShardInfo | null> {
     const metadataDb = await this.ensureInitialized();
     const fileName = basename(dbPath);
-    const row = await metadataDb.get(`SELECT * FROM shards WHERE db_path LIKE '%' || ?`, [
-      fileName,
-    ]);
+    // Stored db_path is always `<scope>s/<basename>` (see createShard/registerExistingShard),
+    // so anchor on the "/" separator. Escape LIKE metacharacters in the filename — otherwise
+    // the "_" in shard names like `user_<hash>_0.db` would match any character.
+    const escaped = fileName.replace(/[\\%_]/g, "\\$&");
+    const row = await metadataDb.get(
+      `SELECT * FROM shards WHERE db_path LIKE '%/' || ? ESCAPE '\\'`,
+      [escaped]
+    );
     if (!row) return null;
     return this.rowToShardInfo(row);
   }
--- a/src/services/user-memory-learning.ts
+++ b/src/services/user-memory-learning.ts
@@ -215,7 +215,7 @@ Rules:
           updatedProfileData
         );
 
-        const validationSummary = applyValidations(
+        const validationSummary = await applyValidations(
           updatedProfileData,
           llmResult,
           existingProfile.id,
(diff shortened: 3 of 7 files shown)


lindixu6-hash (changes requested):
I verified the branch at `e79af2441beda6cef23ce8e5c1cb8015d4766a5a`. The existing gates are green (targeted 22/22, full 408/408, typecheck, Prettier, and production build), but three data-integrity paths remain:

1. **`learning_paths` is still dropped by the normal merge path.** `createProfile`/`updateProfile` now preserve the field only if it reaches them, but `mergeProfileData` reconstructs `{ preferences, patterns, workflows }` and omits `existing.learning_paths`. A real Manager + SQLite reproduction produced `mergedHas=false` and `storedHas=false` after a normal merge/update. Please carry the field through merge and cover the update following `buildLearningPaths` (including automatic cleanup).

2. **Partial malformed AI mappings can still delete originals.** `normalizeAIMapping({ removed: ["pref_0"] })` treats missing `kept`/`merged` as empty but accepts `removed`; rebuild then returns an empty preferences list. This contradicts the stated fail-safe no-op behavior. Please validate the mapping as one contract (all required arrays, known/disjoint IDs, valid merge groups), and make any malformed contract a complete no-op.

3. **`createShard` mutates an existing file before proving the registry slot is new.** `initShardDb` performs `INSERT OR REPLACE` on embedding metadata before the registry INSERT. With an existing inactive index-0 row (the import path calls index 0 when no active shard exists), the unique-key catch returns the old row after its file metadata was already rewritten. My reproduction changed an old 8-dim/old shard to 16-dim/new while `sameId=true` and the row remained inactive, without re-embedding vectors. Please claim/check the registry slot before initializing the target path, or initialize a temporary file and atomically register/rename it.

These are not exposed by the current suite, so each needs a failure-sensitive regression. The other reviewed changes (per-profile cold buffer isolation, awaited evolution ordering, corrupt JSON handling, and escaped path lookup) look directionally sound.

lindixu6-hash:

tickernelz:
Merged into `main` as part of the newest-to-oldest integration batch and scheduled for the next release. The combined tree was verified with frozen installs (root + web), typecheck, production build, and the full test suite (**444 passed / 0 failed / 1 skipped**) across Linux, Windows, macOS Intel and Apple Silicon CI.

tickernelz:
Released in [v2.25.0](https://github.com/tickernelz/opencode-mem/releases/tag/v2.25.0). The combined integration tree was verified with frozen installs (root + web), typecheck, production build, and the full test suite (**444 passed / 0 failed / 1 skipped**) across Linux, Windows, macOS Intel and Apple Silicon, then published to npm.

PR #254, opened 2026-08-16
feat: add named orcarouter memory provider mode

Marc-oss-hub:
## What this PR does

Adds a first-class `orcarouter` `memoryProvider` mode to opencode-mem, backed by a new `OrcaRouterProvider` that extends the existing OpenAI Chat Completions provider.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible model routing gateway that requires namespaced model IDs (for example `orcarouter/auto`, `openai/gpt-5.5`, `deepseek/deepseek-v4-flash`). It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

## How it works

- **Minimal config**: when `memoryApiUrl` and `memoryModel` are omitted they default to `https://api.orcarouter.ai/v1` and `orcarouter/auto` (the gateway's routing alias, which picks a capable upstream model per request). A user only needs:
  ```jsonc
  "memoryProvider": "orcarouter",
  "memoryApiKey": "<OrcaRouter API key>",
  ```
- **Namespacing guard**: a bare model name (e.g. `gpt-4o-mini`) is rejected locally with a hint instead of surfacing a gateway-side `model_not_found` error.
- **Session support**: the provider is tagged `orcarouter` in the AI session store and diagnostics, so memory sessions are distinguishable from `openai-chat`.
- **Config status**: `getAutoCaptureProviderStatus` and `buildMemoryProviderConfig` treat `orcarouter` as a preset — only the API key is required.

The base `OpenAIChatCompletionProvider` gains protected `sessionProviderTag` / `resolveEndpoint` / `resolveModel` hooks so providers can override endpoint, model, and session tagging without duplicating request handling.

## Files changed

- `src/services/ai/providers/orcarouter.ts` — new provider
- `src/services/ai/providers/openai-chat-completion.ts` — extensible hooks
- `src/services/ai/ai-provider-factory.ts` — factory + supported-providers wiring
- `src/services/ai/provider-config.ts` — preset-aware config validation
- `src/services/ai/session/session-types.ts`, `src/types/index.ts` — provider type union
- `src/config.ts` — config template + status for `orcarouter`
- `README.md` — documentation
- `tests/orcarouter-provider.test.ts`, `tests/ai-provider-config.test.ts` — coverage

## Validation

- `bun run typecheck` — clean
- `bun test` — 418 pass; the only 2 failures are pre-existing Windows-environment issues in `tests/onnxruntime-resolve.test.ts` (native DLL load failure and an `EPERM` on `symlinkSync`), unrelated to this change
- `bun run format:check` — clean

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,12 @@ Manual `memoryProvider` modes:
 - `openai-responses`: OpenAI Responses API with function-call output.
 - `anthropic`: Anthropic Messages API with tool use.
 - `minimax`: MiniMax Anthropic Messages-compatible endpoint. Set `memoryApiUrl` to the global endpoint (`https://api.minimax.io`) or the China endpoint (`https://api.minimaxi.com`); the `/anthropic/v1/messages` path and `x-api-key` header are applied automatically. MiniMax text models such as `MiniMax-M3` support the adaptive thinking modes used by this plugin via `memoryExtraParams`.
+- `orcarouter`: OpenAI-compatible model gateway with namespaced model IDs. `memoryApiUrl` and `memoryModel` are optional — they default to `https://api.orcarouter.ai/v1` and `orcarouter/auto` (a routing alias that selects a capable model per request). If you set `memoryModel`, use a namespaced ID such as `openai/gpt-5.5` or `deepseek/deepseek-v4-flash`; OrcaRouter rejects bare model names. Example:
+  ```jsonc
+  "memoryProvider": "orcarouter",
+  "memoryApiKey": "<OrcaRouter API key>",
+  ```
+  [OrcaRouter](https://www.orcarouter.ai) also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.
 
 Troubleshooting:
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -44,7 +44,7 @@ interface OpenCodeMemConfig {
   autoCaptureMaxRetries?: number;
   autoCaptureMaxContextBytes?: number;
   autoCaptureLanguage?: string;
-  memoryProvider?: "openai-chat" | "openai-responses" | "anthropic" | "minimax";
+  memoryProvider?: "openai-chat" | "openai-responses" | "anthropic" | "minimax" | "orcarouter";
   memoryModel?: string;
   memoryApiUrl?: string;
   memoryApiKey?: string;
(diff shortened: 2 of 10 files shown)


lindixu6-hash (changes requested):
The provider unit coverage is thorough, and the submitted branch passes typecheck, formatting, production build, and the full suite (`420 pass`, `0 fail`). However, the advertised API-key-only configuration is not reachable from the two primary manual fallback paths.

`getAutoCaptureProviderStatus()` reports `{ ready: true, mode: "manual" }` for `memoryProvider: "orcarouter"` plus only `memoryApiKey`, and `buildMemoryProviderConfig()` accepts it. But `generateSummary()` in `src/services/auto-capture.ts` still throws before the factory whenever `CONFIG.memoryModel` or `CONFIG.memoryApiUrl` is absent. `analyzeUserProfile()` in `src/services/user-memory-learning.ts` has the same pre-factory guard. Several direct profile fallback paths also remain gated on both fields.

I reproduced this on commit `bdabf8c` with a temporary integration test through `performAutoCapture()`: status was ready/manual, but `providerCreated=false`, `saved=false`, and the prompt failure count became 1. Temporarily allowing the OrcaRouter preset through that guard made the same test pass without changing its mocks. The temporary test and mutation were then removed; the review worktree is clean.

Please centralize the preset readiness/default resolution so the actual auto-capture and user-profile-learning entry points honor the same contract, and add failure-sensitive integration coverage through those entry points, not only provider/config unit tests.

Separately, the README statement that the vendor provides gateway-level “zero-trust” screening and governs every tool call on a default-deny basis is not established by this repository or its tests. Since this is a vendor-authored preset, please keep the project documentation neutral: disclose that prompts/responses are sent to the external gateway, link official vendor documentation for optional security claims, and avoid presenting those claims as a verified property of opencode-mem.

Once the real entry paths and documentation boundary are corrected, the provider subclassing/session-tag approach is reasonable.

lindixu6-hash:

tickernelz:
Merged into `main` as part of the newest-to-oldest integration batch and scheduled for the next release. The combined tree was verified with frozen installs (root + web), typecheck, production build, and the full test suite (**444 passed / 0 failed / 1 skipped**) across Linux, Windows, macOS Intel and Apple Silicon CI.

tickernelz:
Released in [v2.25.0](https://github.com/tickernelz/opencode-mem/releases/tag/v2.25.0). The combined integration tree was verified with frozen installs (root + web), typecheck, production build, and the full test suite (**444 passed / 0 failed / 1 skipped**) across Linux, Windows, macOS Intel and Apple Silicon, then published to npm.

PR #257, opened 2026-08-17
fix: preserve vector-first ANN query plan

### LINKED ISSUE #247: [Bug]: memory search takes ~8s when filtering by container_tag (vector_top_k + JOIN bad query plan)
### Environment
- opencode-mem 2.24.3
- Node.js v26.5.0, Windows
- Local embedding (Xenova/nomic-embed-text-v1), local libsql shard (no network)

### Symptom
Searching memories in the Web UI hangs for ~40s total; the frontend eventually aborts the fetch with `Error: signal is aborted without reason`. Profiling shows the bottleneck is `searchKind()` in `dist/services/turso/vector-search.js`: a single ANN search against a 337-row shard takes **~8 seconds** (3 runs: 7982/8113/8064 ms) instead of tens of milliseconds.

### Root cause
`searchKind()` combines `vector_top_k()` with `INNER JOIN memories m ON m.rowid = v.id` and an optional `container_tag` filter:

```sql
SELECT m.id AS id, vector_distance_cos(m.content_vec, vector32(?)) AS dist
FROM vector_top_k('memories_content_vec_idx', vector32(?), ?) AS v
JOIN memories m ON m.rowid = v.id
WHERE m.content_vec IS NOT NULL AND m.container_tag = ?
```
EXPLAIN QUERY PLAN shows SQLite's planner drives from the `memories` table (using `idx_container_tag`) and then evaluates the `vector_top_k` virtual table **once per memory row** — turning one ANN lookup into N nested-loop evaluations (337 rows = 337 ANN calls):

```
SEARCH m USING INDEX idx_container_tag (container_tag=?)
SCAN v VIRTUAL TABLE INDEX 1:
```
### Fix
All timings below were benchmarked on the **unmodified codebase** by running the candidate SQL directly against the database in a test script; the source was not modified during investigation.

Changing `JOIN` to `CROSS JOIN` forces the intended TVF-first plan and produces the correct result in **~28 ms** (re-run: 29/29/29 ms, 199 rows):

```
SCAN v VIRTUAL TABLE INDEX 1:
SEARCH m USING INTEGER PRIMARY KEY (rowid=?)
```
Benchmark (same query vector, 337-row shard):

| Variant | Time |
|---|---|
| INNER JOIN (current code path) | 8409 / 8534 / 8881 ms |
| CROSS JOIN (suggested fix) | 29 / 29 / 29 ms |
| Two-step (top_k then rowid IN) | 29 ms (content) + 28 ms (tags) |

Semantics are unchanged (`WHERE m.vector IS NOT NULL` / `container_tag` filters still apply).

Suggested patch in `dist/services/turso/vector-search.js` `searchKind()` (both branches):

```diff
-          JOIN memories m ON m.rowid = v.id
+          CROSS JOIN memories m ON m.rowid = v.id
```
Alternative: two-step query (top_k first, then `SELECT ... WHERE rowid IN (...)`) also works, but CROSS JOIN is a one-word change and is therefore preferred.

lindixu6-hash:
## Summary

- use `CROSS JOIN` in both `vector_top_k` query paths so SQLite keeps the virtual table before `memories`
- preserve the existing filtered and unfiltered result semantics
- add a regression test covering both SQL branches

## Why

Issue #247 shows SQLite can reorder the current inner join, drive from `memories`, and evaluate `vector_top_k` once per matching row. `CROSS JOIN` acts as the planner barrier proposed in the issue while keeping the join predicate and filters unchanged.

Closes #247.

## Verification

- `bun run typecheck`
- `bun run format:check`
- `bun run build`
- `bun test` (`408 pass`, `0 fail`)

### DIFF
--- a/src/services/turso/vector-search.ts
+++ b/src/services/turso/vector-search.ts
@@ -236,13 +236,13 @@ export class TursoVectorSearch {
           ? `
           SELECT m.id AS id, vector_distance_cos(m.${columnName}, vector32(?)) AS dist
           FROM vector_top_k('${indexName}', vector32(?), ?) AS v
-          JOIN memories m ON m.rowid = v.id
+          CROSS JOIN memories m ON m.rowid = v.id
           WHERE m.${columnName} IS NOT NULL
         `
           : `
           SELECT m.id AS id, vector_distance_cos(m.${columnName}, vector32(?)) AS dist
           FROM vector_top_k('${indexName}', vector32(?), ?) AS v
-          JOIN memories m ON m.rowid = v.id
+          CROSS JOIN memories m ON m.rowid = v.id
           WHERE m.${columnName} IS NOT NULL AND m.container_tag = ?
         `,
         containerTag === "" ? [queryJson, queryJson, k] : [queryJson, queryJson, k, containerTag]
--- a/tests/turso-vector-search.test.ts
+++ b/tests/turso-vector-search.test.ts
@@ -68,4 +68,36 @@ describe("turso vector search", () => {
     );
     expect(limitedResults).toHaveLength(1);
   });
+
+  it("keeps vector_top_k before memories for filtered and unfiltered ANN queries", async () => {
+    const observedSql: string[] = [];
+    const db = {
+      all: async (sql: string) => {
+        observedSql.push(sql);
+        return [];
+      },
+    };
+    const { tursoVectorSearch } = await import("../src/services/turso/vector-search.js");
+    const search = tursoVectorSearch as unknown as {
+      searchKind(
+        database: typeof db,
+        queryJson: string,
+        k: number,
+        containerTag: string,
+        indexName: string,
+        columnName: string
+      ): Promise<Array<{ id: string; similarity: number }>>;
+    };
+
+    await search.searchKind(db, "[1,0]", 10, "", "memories_vec_idx", "vector");
+    await search.searchKind(db, "[1,0]", 10, "opencode_project_test", "memories_vec_idx", "vector");
+
+    expect(observedSql).toHaveLength(2);
+    for (const sql of observedSql) {
+      expect(sql).toContain("FROM vector_top_k");
+      expect(sql).toContain("CROSS JOIN memories m ON m.rowid = v.id");
+      expect(sql.indexOf("vector_top_k")).toBeLessThan(sql.indexOf("memories m"));
+    }
+    expect(observedSql[1]).toContain("m.container_tag = ?");
+  });
 });


lindixu6-hash:
Remote validation for the exact PR commit `c12b4e3334ca80e58acd677ad28f1f61fc3cb10d`: [Platform Package Smoke run 32068871957](https://github.com/lindixu6-hash/opencode-mem/actions/runs/32068871957) passed all six jobs (Ubuntu, Windows, macOS Apple Silicon 15/26, and macOS Intel 15/26), including install, typecheck, build, the full test suite, and package-install smoke.

For transparency, the first Windows attempt hit the existing 5-second cleanup timeout in `turso-reembed-migration.test.ts`; the vector-search tests passed in that attempt, and the Windows rerun completed fully green without code changes.

lindixu6-hash:
Hi @EyJunge1, when convenient, could you independently review this PR and the related diagnostic fix in #258? You are the recent merger for the Turso, auto-capture, dependency, and v2.24.x release work, so these two paths align with your recent context.

Both PRs are limited to two files, remain mergeable, and have exact-SHA six-platform package-smoke evidence. I will not self-approve or merge either change.

EyJunge1:
Thanks for the ping! Honestly, I'm no longer maintaining this project and don't have the time to give this PR a proper review. Better to loop in one of the active maintainers who can do it justice. Appreciate your understanding.

lindixu6-hash:
Fresh validation after updating this branch to cumulative latest `main`: exact PR head `70eb41ce7998812e06cc626ee6376443bc2ea1a4` passed [Platform Package Smoke run 32281742023](https://github.com/tickernelz/opencode-mem/actions/runs/32281742023) on Ubuntu, Windows, macOS 15/26 Apple Silicon, and macOS 15/26 Intel, plus GitGuardian. The PR remains independently review-gated; I will not self-approve or merge it.

NaNomicon:
https://github.com/tickernelz/opencode-mem/pull/258#issuecomment-5367974637

NaNomicon:
## Review — PR #257: "fix: preserve vector-first ANN query plan"

**Verdict: SHIP.** Verified the planner-barrier fix against real libSQL. Minimal, correct, no blast radius.

**The fix**: `JOIN` → `CROSS JOIN` in both `searchKind` SQL branches (`src/services/turso/vector-search.ts:239,245`). CROSS JOIN is SQLite's documented optimizer barrier — it forces `vector_top_k` (virtual table) to stay the driving table so it is evaluated once, not once per matching `memories` row. Join predicate `m.rowid = v.id` and all WHERE filters unchanged.

### What is actually fixed (the interesting part)

Only the **filtered** branch was broken. Confirmed with `EXPLAIN QUERY PLAN` against real libSQL (v0.5.29):

```
OLD filtered:  SEARCH m USING INDEX idx_container_tag   <- memories drives, vector_top_k per row (bug #247)
NEW filtered:  SCAN v VIRTUAL TABLE ...                 <- vector_top_k drives, fixed
```

The unfiltered branch already picked `v`-first naturally — CROSS JOIN is a no-op there but harmless.

### Why it is safe

CROSS JOIN with an ON clause is semantically identical to INNER JOIN — it only constrains planner join order, not the result set. Verified:

- **Real-execution** test (`inserts and searches memories with native vector index`) drives the modified query against a real SQLite shard — passes, correct ranking.
- **Row-count equivalence** at production k (128 filtered / 32 unfiltered): old and new return identical result sets.
- **Graceful fallback intact**: CROSS JOIN against a missing index still throws into `exactScanKind` fallback.
- **No twin-copy miss**: only 2 `JOIN memories` in the codebase, both changed.
- `m.rowid = v.id` is correct — `memories` has a `TEXT PRIMARY KEY`, so `rowid` is the physical rowid (not aliased to the PK), and `vector_top_k` returns those integer rowids. Integer-to-integer join, unaffected by the keyword change.

### On the test

The added test (`tests/turso-vector-search.test.ts`) mocks `db.all` and asserts the **emitted SQL shape** — it guards against revert but does not execute SQLite. That is fine here because the existing real-SQLite test already exercises the modified query. 2/2 pass.

**Optional hardening** (not required): add an `EXPLAIN QUERY PLAN` assertion to the real-SQLite test checking the filtered plan drives from `vector_top_k` — would catch a future libSQL/SQLite version that stops honoring the CROSS-JOIN barrier. Belt-and-suspenders only.

### Merge gate

- Correct, minimal, fail-safe (graceful fallback preserved)
- Real-SQLite execution verified
- Test guards the revert
- No breakage beyond the diff (verified callers, triggers, fallback)

**Merge as-is.** Closes #247 properly. (Verified via static code + EXPLAIN QUERY PLAN + real execution; live behavior across SQLite versions relies on the CROSS-JOIN planner-barrier semantics, which are standard SQLite.)

lindixu6-hash:
Thanks for the thorough independent review and the real-libSQL plan verification. I agree the `EXPLAIN QUERY PLAN` assertion is useful optional hardening, but keeping it out preserves this PR's narrow scope.

All checks remain green. Since this is my PR, I will not self-merge it; please merge when convenient.

Issue #253, opened 2026-08-15
Thinking mode does not support this tool_choice


augustwy:
模型执行结束后报错
<img width="532" height="153" alt="Image" src="https://github.com/user-attachments/assets/0e58c4f8-2da3-449c-82b0-335a2b133ba8" />

<img width="561" height="136" alt="Image" src="https://github.com/user-attachments/assets/c07f4a6a-6865-499a-971c-00c3dff2b6e1" />

opencode 模型配置如下
"opencodeProvider": "opencode-go",
 "opencodeModel": "deepseek-v4-flash"

lindixu6-hash:
我复核了两张截图和当前调用链，这里实际有两层问题：

1. 底层真实错误是 `opencode-go / deepseek-v4-flash` 的 thinking 模式拒绝 forced `tool_choice`。`opencode-mem` 走的是 OpenCode `session.prompt` + `format: json_schema`，provider wire payload 由 OpenCode 生成；本仓库的 manual `OpenAIChatCompletionProvider.tool_choice` 不参与这条路径，因此不应在这里硬编码 DeepSeek 模型名单或伪造 provider 参数。
2. 当前插件捕获底层错误后，无条件尝试可选 External API fallback；当用户没有配置 manual fallback 时，又用 `External API not configured for auto-capture` 覆盖了真实根因，所以最终 toast 误导。

我已提交 #258 修复第 2 层：没有完整 manual fallback 时保留原始 OpenCode provider 错误；若确实配置了 `memoryModel` + `memoryApiUrl`，原 fallback 行为不变。入口级回归证明 toast 会显示 `Thinking mode does not support ...`，不会再显示误导性的 External API 错误。

本地验证：目标 2/2、完整 408/408、typecheck、format、production build 全绿；mutation check 移除实现后会精确恢复原截图错误并使回归失败。

第 1 层需要 OpenCode/provider compat 修复：同一 `opencode-go/deepseek-v4-flash` + OpenAI Responses 错误已被其他客户端独立复现，正确策略是在 thinking 模型上省略 forced named selector、保留工具广告，而不是由插件自行重写供应商 payload。当前绕行是升级到包含该兼容策略的 OpenCode/provider 版本，或为 `opencodeModel` 选择支持 structured output 的非-thinking 模型。

#258 不声称修改上游 `tool_choice`，只确保用户能看到可操作的真实错误。

lindixu6-hash:
[...]

PR #258, opened 2026-08-17
fix: preserve opencode auto-capture errors

lindixu6-hash:
## Summary

- keep the original OpenCode structured-output failure when no manual External API fallback is configured
- preserve the existing fallback behavior when `memoryModel` and `memoryApiUrl` are available
- add an entry-point regression proving the user-visible toast retains the provider root cause and releases the prompt claim

## Why

Issue #253 shows the underlying provider error (`Thinking mode does not support this tool_choice`) followed by a misleading final toast (`External API not configured for auto-capture`). The OpenCode path currently catches every provider error, then overwrites it when the optional manual fallback is absent.

This patch does not alter OpenCode/provider wire payloads or hard-code a model list. Forced `tool_choice` compatibility belongs to the upstream provider layer; this repository should preserve that actionable error instead of masking it.

Refs #253.

## Verification

- `bun test tests/auto-capture.test.ts` (2/2)
- full `bun test` (408/408)
- `bun run typecheck`
- `bun run format:check`
- `bun run build`
- mutation check: removing only the six implementation lines restores `External API not configured for auto-capture` and makes the new regression fail

### DIFF
--- a/src/services/auto-capture.ts
+++ b/src/services/auto-capture.ts
@@ -484,6 +484,8 @@ async function generateSummary(
   userPrompt: string,
   prompt?: { id?: string; providerId: string | null; modelId: string | null }
 ): Promise<{ summary: string; type: string; tags: string[] } | null> {
+  let opencodeProviderError: unknown;
+
   // Opencode provider path (when opencodeProvider + opencodeModel configured)
   if (CONFIG.opencodeProvider && CONFIG.opencodeModel) {
     try {
@@ -572,6 +574,7 @@ CAPTURE if: code changed, bug fixed, feature added, decision made`;
         tags: (result.tags || []).map((t: string) => t.toLowerCase().trim()),
       };
     } catch (e) {
+      opencodeProviderError = e;
       log("auto-capture: opencode provider failed, falling back to external API", {
         error: String(e),
       });
@@ -580,6 +583,9 @@ CAPTURE if: code changed, bug fixed, feature added, decision made`;
 
   // Existing manual config path
   if (!CONFIG.memoryModel || !CONFIG.memoryApiUrl) {
+    if (opencodeProviderError) {
+      throw opencodeProviderError;
+    }
     throw new Error("External API not configured for auto-capture");
   }
 
--- a/tests/auto-capture.test.ts
+++ b/tests/auto-capture.test.ts
@@ -219,6 +219,122 @@ console.log(
   };
 }
 
+function runProviderFailureScenario() {
+  const dir = mkdtempSync(join(tmpdir(), "opencode-mem-auto-capture-error-"));
+  tempDirs.push(dir);
+  const scriptPath = join(dir, "scenario.mjs");
+
+  const script = `
+import { mock } from "bun:test";
+
+const toasts = [];
+let failedAttempts = 0;
+let released = false;
+
+mock.module(${JSON.stringify(configUrl)}, () => ({
+  CONFIG: {
+    autoCaptureMaxRetries: 1,
+    autoCaptureProviderStatus: { ready: true, mode: "opencode", issues: [] },
+    autoCaptureLanguage: "en",
+    opencodeProvider: "opencode-go",
+    opencodeModel: "deepseek-v4-flash",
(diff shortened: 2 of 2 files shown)


lindixu6-hash:
Remote validation for exact commit `6fe66aa87657727f8f7fd39a1079315c92f3c0fb`: [Platform Package Smoke run 32071720465](https://github.com/lindixu6-hash/opencode-mem/actions/runs/32071720465) passed all six jobs (Ubuntu, Windows, macOS 15/26 Apple Silicon, and macOS 15/26 Intel). Each job completed frozen install, typecheck, production build, the full test suite, package creation, package installation, and native/runtime smoke checks.

lindixu6-hash:
Fresh validation after updating this branch to cumulative latest `main`: exact PR head `6ff867ad447eb79cb2607ca651ebeac9c0147f2f` passed [Platform Package Smoke run 32281747641](https://github.com/tickernelz/opencode-mem/actions/runs/32281747641) on Ubuntu, Windows, macOS 15/26 Apple Silicon, and macOS 15/26 Intel, plus GitGuardian. The PR remains independently review-gated; I will not self-approve or merge it.

NaNomicon:
Happy to take a look at the PR, we're (at least I am) honestly not using the plugin any more so it'll be awkward to maintain it, feel free to take control after the 2 PRs

NaNomicon (commented):
## Review: preserve opencode auto-capture errors

**Call: SHIP.** Change is correct, minimal, tested, fail-closed. Nothing blocks merge.

### What the change does
Six lines in `generateSummary` (`src/services/auto-capture.ts`). When the opencode structured-output path throws and no manual external-API fallback is configured, the real provider error is now rethrown and surfaced to the user instead of being masked by the generic `"External API not configured for auto-capture"` toast. When `memoryModel` + `memoryApiUrl` are present, the manual fallback runs exactly as before — the provider error stays in the logs.

### Why it is safe
- **Scoped tight.** The rethrow only fires in the no-fallback branch. The manual path is unchanged.
- **Failure direction is closed.** Error now surfaces instead of being swallowed. The only remaining swallow is the intended one: opencode fails → manual fallback runs → manual error reported, first error in logs. That is a fallback chain, not a mask.
- **Error type is safe.** Every throw site in `generateStructuredOutput` produces `Error` (verified: `new Error(...)`, zod `ZodError`, fetch `TypeError`). Even if a non-Error slipped through, all three catch points in the caller use `instanceof Error ? .message : String(e)` — no unhandled rejection, toast always renders.

### Test is real, not decorative
`runProviderFailureScenario()` spawns a real bun subprocess and asserts `failedAttempts=1`, `released=true`, exactly one toast, message containing the provider root cause, and not the old mask. Each assertion eliminates a distinct wrong path — early returns, silent swallows, claim failure all break one of them. Genuine regression test, not an inert mock.

### Not blockers
- `user-memory-learning.ts:697` (`analyzeUserProfile`) carries the same error-mask in the opencode path — same bug class, different service. It deserves the same fix on its own PR; this ones contract is auto-capture. Recommend filing a follow-up so it does not rot.
- Stale log at `src/services/auto-capture.ts:578` (`"falling back to external API"` when no fallback exists) — pre-existing, slightly more visible now. Nit, fix on next touch.
- Body claims `408/408`; current main is `411/411`. Stale count, CI green, immaterial.

### Scope note
This review is **static analysis only** — code read, test logic traced, error paths checked. Actual runtime behavior (subprocess spawn, toast rendering, provider interaction) relies on the PR author owning the change; I did not run the full suite against a live provider.

Merge. File the `user-memory-learning` twin as a follow-up.

lindixu6-hash:
Thanks for the detailed review and the maintenance handoff. I'm happy to take a bounded lane covering Turso/vector-search reliability, auto-capture/profile-learning reliability, issue triage, and CI/release readiness. I'll continue requiring independent review for my own changes.

I filed #265 for the `user-memory-learning` error-mask and stale fallback-log follow-up so #258 can remain narrowly scoped. All checks remain green; please merge #257 and #258 when convenient.

Issue #251, opened 2026-08-13
[Bug] Web server takeover loop on Windows when orphaned LISTEN socket holds the port


skywalker-35:
## Summary

On Windows, when the opencode process dies abnormally (crash / killed), its web server socket can remain in the kernel TCP table as an **orphaned LISTEN socket** (the owning PID no longer exists). On the next start, `WebServer._start()` hits `EADDRINUSE`, falls into `startHealthCheckLoop()`, and gets stuck in a **takeover death loop** — the web UI never comes back until Windows reclaims the socket (observed: 2–4 hours) or the machine reboots.

## Environment

- OS: Windows 11
- opencode-mem: 2.24.3
- Runtime: Bun (bundled with opencode)

## Symptoms

```
netstat -ano                     → TCP 127.0.0.1:4748 LISTENING PID 20748
Get-Process -Id 20748            → (nothing — the process is dead)
Test-NetConnection :4748         → TcpTestSucceeded: False
Invoke-WebRequest :4748          → connection refused
opencode-mem log                 → MCP backend works fine (auto-capture keeps
                                   persisting), but zero web-server entries after
                                   "Embedding model warmed up"
```

`healthCheckLoop` → `checkServerAvailable()` (fetch timeout) → `attemptTakeover()` → bind `EADDRINUSE` → loop every 5 s, forever.

## Root cause

The takeover mechanism introduced in #25 / PR #26 uses an **HTTP probe** (`fetch("/api/health")`) in `checkServerAvailable()` to decide whether the port holder is alive. That encodes the assumption:

> port occupied = another healthy opencode instance owns it

An orphaned Windows socket breaks that assumption: it holds the port (so `bind` fails with `EADDRINUSE`) but nothing answers HTTP (so the fetch always fails). Result: probe says "not available" → takeover fires → bind fails → loop.

There are actually **two flavors** of orphaned socket on Windows:

| Type | TCP connect | HTTP fetch | bind | Origin |
|---|---|---|---|---|
| A (like #32) | ✅ accepted | ❌ never responds | EADDRINUSE | Bun worker thread residue |
| B | ❌ RST (refused) | ❌ refused | EADDRINUSE | Kernel TCP-table residue after abnormal process exit |

PR #26 fixed takeover but probes via HTTP only; PR #33 fixed the Bun worker but doesn't cover orphaned sockets. Together they leave this Windows-specific gap.

## Suggested fixes

[...]

PR #259, opened 2026-08-19
fix: vector search perf regression + Windows web server takeover loop

petra-dot:
Fixes two reported bugs.

## #247: memory search takes ~8s when filtering by container_tag
`searchKind()` combined `vector_top_k()` with `INNER JOIN memories m ON m.rowid = v.id`. SQLite's planner drove from the `memories` table and evaluated the ANN index once per row (337 rows = 337 ANN calls). `CROSS JOIN` forces the table-function-first plan; semantics unchanged. Benchmarked in the issue: 8.4s -> ~29ms. Added a regression test asserting the query plan scans `vector_top_k` before the memories lookup.

## #251: Web server takeover loop on Windows when an orphaned LISTEN socket holds the port
After a crash, an orphaned socket holds the configured port (EADDRINUSE) but answers no HTTP, so `attemptTakeover()` retried forever every 5s. After 3 consecutive failed takeovers the server now falls back to `port+1` (bounded at `port+10`) and the takeover toast surfaces the actual URL. Added a unit test for the port-fallback policy.

## Verification
- `tsc --noEmit` clean
- `bun run build` succeeds
- Full test suite: 405 pass / 4 fail, identical failure set on clean `main` (pre-existing: onnxruntime-resolve shim, plugin-loader contract timeout, turso-shard recreate timeout, config parallel flake). No new failures.
- Prettier clean on all touched files

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -326,8 +326,23 @@ export const OpenCodeMemPlugin: Plugin = async (ctx: PluginInput) => {
               .showToast({
                 body: {
                   title: "Memory Explorer",
-                  message: `Web UI started at ${webServer?.getUrl() ?? url}`,
+                  message: "Took over web server ownership",
                   variant: "success",
+                  duration: 3000,
+                },
+              })
+              .catch(() => {});
+          }
+        });
+
+        webServer.setOnPortsExhaustedCallback(() => {
+          if (ctx.client?.tui) {
+            ctx.client.tui
+              .showToast({
+                body: {
+                  title: "Memory Explorer",
+                  message: `Web UI unavailable: ports ${CONFIG.webServerPort}-${CONFIG.webServerPort + 10} are held by non-responsive processes`,
+                  variant: "error",
                   duration: 5000,
                 },
               })
--- a/src/index.ts
+++ b/src/index.ts
@@ -326,9 +326,9 @@ export const OpenCodeMemPlugin: Plugin = async (ctx: PluginInput) => {
               .showToast({
                 body: {
                   title: "Memory Explorer",
-                  message: "Took over web server ownership",
+                  message: `Web UI started at ${webServer?.getUrl() ?? url}`,
                   variant: "success",
-                  duration: 3000,
+                  duration: 5000,
                 },
               })
               .catch(() => {});
--- a/src/services/web-server.ts
+++ b/src/services/web-server.ts
@@ -156,6 +156,24 @@ function serveFetch(opts: {
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
+/**
+ * Port fallback policy for the takeover loop. A port that fails to bind AND
+ * answers no HTTP is treated as orphaned Windows kernel residue; after
+ * `minFailedTakeovers` consecutive failed takeovers the web server moves to
+ * `currentPort + 1`, bounded by `maxFallbackPort`.
+ */
+export function nextFallbackPort(
+  currentPort: number,
+  failedTakeovers: number,
+  maxFallbackPort: number,
+  minFailedTakeovers: number = 3
+): number {
+  if (failedTakeovers < minFailedTakeovers || currentPort >= maxFallbackPort) {
+    return currentPort;
+  }
+  return currentPort + 1;
+}
+
 interface WebServerConfig {
   port: number;
   host: string;
@@ -171,9 +189,12 @@ export class WebServer {
   private startPromise: Promise<void> | null = null;
   private healthCheckInterval: NodeJS.Timeout | null = null;
   private onTakeoverCallback: (() => Promise<void>) | null = null;
+  private takeoverFailures: number = 0;
+  private readonly maxFallbackPort: number;
 
   constructor(config: WebServerConfig) {
     this.config = config;
+    this.maxFallbackPort = config.port + 10;
   }
 
   setOnTakeoverCallback(callback: () => Promise<void>): void {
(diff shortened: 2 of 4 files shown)


petra-dot:
Note: this PR overlaps with #257, which makes the same CROSS JOIN change for #247. Either merge order works — the hunk is identical. This PR additionally includes the #251 web-server takeover fix and regression tests for both, so you can drop #257 if preferred, or take #257 and I'll rebase this one down to just the #251 fix.

lindixu6-hash (commented):
Reviewed the #247 overlap and the new #251 path. The `CROSS JOIN` implementation matches #257 and the query-plan assertion is useful. One coverage difference to preserve if this PR supersedes #257: #257 exercises both `searchKind()` SQL branches (empty and non-empty `containerTag`), while this regression executes only the filtered query. Please retain coverage for both generated branches in whichever PR lands.

For #251, `nextFallbackPort()` is well bounded, but the current test exercises only that pure helper. The behavior depends on the surrounding takeover state machine: repeated `_start()`/`EADDRINUSE`, health-loop restart, owner transition, and callback URL after `config.port` changes. A focused integration-style test around those transitions would make the Windows fix substantially safer.

I attempted the two targeted tests locally, but the isolated worktree dependency install stalled and the partial environment could not resolve `@libsql/client`, so I am not claiming runtime verification from that attempt.

petra-dot:
Addressed both review points in d12f0d0:

1. **#247 both branches** — the plan assertion now runs for both `searchKind()` SQL variants: "drives vector_top_k before the memories join when filtering by container_tag" and "...without a container tag". The empty-tag branch (`WHERE m.vector IS NOT NULL`, no container predicate) is covered alongside the filtered one.

2. **#251 takeover state machine** — new integration-style test "becomes owner on a neighbor port after repeated failed takeovers": a real listener holds the port (EADDRINUSE on every bind), `checkServerAvailable` reports unavailable, and three direct `attemptTakeover()` cycles run the real state machine — first two leave `isServerOwner()` false on the original port, the third bumps to the neighbor port, flips ownership, fires the takeover callback once, and `getUrl()` reports the new port with `/api/health` answering on it.

Verification: both test files 6 pass, `tsc --noEmit` clean, full suite 408 pass / 3 fail (same pre-existing flakes as clean main: onnxruntime shim, plugin-loader timeout, turso-shard-recreate timeout), prettier clean.

petra-dot:
Diagnosing the windows-latest failure on this run (no changes made to the diff).

**Failure:** `turso legacy migrator dimension preflight > preserves legacy dimensions` — a 10s `afterEach` hook timeout.

**Not caused by this PR:** the branch diff touches only `src/services/turso/vector-search.ts`, `src/services/web-server.ts`, `src/index.ts`, and two test files. The migrator and the shared test-cleanup path are untouched. The same test passes in ~145ms locally on Windows, and the other 5 OS jobs in this matrix passed.

**Mechanism:** the timeout is in `cleanupTursoTestDirectory()` (`tests/turso-test-utils.ts`), which runs `withSqliteFileLockRetry(() => rmSync(...))` after closing Turso handles. On a busy Windows runner, an SQLite file handle can stay locked past the 10s hook window. This is the same flake class as the `turso-shard-recreate` hook timeout seen in earlier local runs — it lands on a different turso test each time.

If you can, re-run the failed job (`gh run rerun <id> --failed`); it should go green. Happy to make the cleanup hook more robust (e.g. raise the hook timeout or degrade the file-lock retry) as a separate PR if this keeps flaking.

lindixu6-hash (changes requested):
The follow-up at `d12f0d0` resolves both points from my first review: both `searchKind()` SQL branches now assert the vector-first plan, and the integration-style test exercises repeated bind failures through owner transition, callback, active URL, and a live health response. I independently ran the two target files: 6/6 tests passed, and `bun run typecheck` passed. The command itself returned non-zero only when the TRAE sandbox blocked cleanup of macOS `/var/folders/...` temp directories after the tests completed.

I also inspected the Windows failure and triggered a failed-job rerun. Both new vector tests and the takeover integration test passed on Windows; the failure was the existing `turso-migrate-dims-preflight` SQLite `EBUSY`/10-second hook timeout in untouched code.

Two state-machine gaps still need changes before this is safe to merge:

1. `takeoverFailures` is described as consecutive, but a successful availability probe does not reset it. Sequence: two failed takeovers, the original owner recovers and serves healthy checks, then dies later; the next single failure advances to a fallback port because the stale count is still 2. Reset the counter when the post-jitter `checkServerAvailable()` succeeds (and after a successful bind), and add a regression for failure -> recovery -> failure.

2. Candidate exhaustion is bounded numerically but not behaviorally. At `maxFallbackPort`, `nextFallbackPort()` returns the same port forever and `startHealthCheckLoop()` retries every five seconds indefinitely. This recreates the loop #251 is intended to eliminate when all 11 candidates are unavailable. Add an explicit exhausted terminal/degraded state with one actionable log/callback (no endless retries), plus coverage for every candidate occupied/unavailable.

The existing health probe also accepts any HTTP 2xx as an opencode-mem owner. Since fallback expands adoption across 10 neighboring ports, validating the `/api/health` JSON shape (`success: true`, `status: "ok"`) would prevent convergence on an unrelated local service; please include that in the candidate handling or explain a stronger existing identity guarantee.

petra-dot:
All three points addressed in af5dbaf:

1. **Stale takeover counter** — `takeoverFailures` now resets in both recovery paths: when the post-jitter `checkServerAvailable()` succeeds (owner recovered, we go passive again) and after a successful bind. Regression test "resets the takeover failure counter when the owner recovers" drives fail → fail → recover → die and asserts the next single failure does NOT bump the port.

2. **Port exhaustion** — at `maxFallbackPort`, once `takeoverFailures` reaches the per-candidate threshold, the health loop is stopped and a terminal degraded state fires exactly once: a log plus an optional `setOnPortsExhaustedCallback` (wired in `index.ts` to a toast with the candidate range). No more 5s retry loop. Test "enters a terminal state when every candidate port is unavailable" asserts the loop is cleared and the callback fires once across subsequent attempts.

3. **Owner identity** — `checkServerAvailable` now parses the response and requires our API envelope: `/api/health` needs `success === true` and `status === "ok"`; the token path (`/api/stats`) needs `success === true`. A 2xx HTML page or a `{ success: false }` body from an unrelated service on a fallback port is treated as unavailable. Test "does not treat a 2xx from an unrelated service as an opencode-mem owner" covers both non-JSON and wrong-shape responses.

Verification: web-server-health.test.ts 6 pass, `tsc --noEmit` clean, prettier clean, full suite 411 pass / 3 fail — the remaining failures are the pre-existing environment flakes (onnxruntime shim fixture, and the Windows SQLite file-lock hook timeout that you re-ran green), none in this PR's code.

lindixu6-hash (approved):
The `af5dbaf` follow-up resolves all three requested state-machine issues:

- recovery resets the consecutive-failure counter, including a fail/fail/recover/fail regression;
- final-candidate exhaustion stops the health loop and emits one bounded degraded-state callback;
- health ownership now requires the opencode-mem JSON envelope instead of accepting arbitrary HTTP 2xx responses.

Independent verification on exact head `af5dbaf640e900cbb72074d1cb4494c0f3c67000`:

- `tests/turso-vector-search.test.ts` + `tests/web-server-health.test.ts`: 9 pass, 0 fail, 36 assertions;
- `bun run typecheck`: pass;
- Prettier check on all five changed source/test files: pass;
- upstream six-platform package-smoke run 32288060862: Ubuntu, Windows, macOS 15/26 Intel, and macOS 15/26 Apple Silicon all pass.

The local test command reported non-zero only after all 9 tests passed because the TRAE sandbox blocked cleanup of temporary test directories; this is not a test or implementation failure.

Approved. If this PR lands first, its two-branch ANN coverage supersedes the overlapping query-plan portion of #257; #257 should then be closed or rebased rather than merged independently.

lindixu6-hash:
@tickernelz, this is now approved at exact head `af5dbaf`: 9/9 focused tests, typecheck, Prettier, and all six package-smoke platforms pass.

One owner decision remains because its ANN hunk overlaps my authored #257. I will not use Write access to choose a merge order that benefits my own PR. Please choose either:

1. independently review/merge #257 first, then have #259 rebased to the Windows takeover changes; or
2. merge #259 first, after which I will close #257 as transparently superseded rather than duplicate the change.

#258 remains separate and independently review-gated. I will not self-approve or self-merge it.

PR #260, opened 2026-08-19
fix: keep orphaned linked memories visible in the timeline

petra-dot:
## Bug

`handleListMemories` (the Web UI timeline endpoint) grouped memory-prompt pairs by id and then dropped any pair missing one side. A memory whose linked prompt was deleted (`DELETE /api/prompts/:id` without cascade), expired, or never captured, vanished from `/api/memories` even though it still existed in the shard and was still reachable via `/api/search`. Same for prompts whose linked memory was gone.

## Root cause

In `handleListMemories`:

```ts
const pairs = Array.from(linkedPairs.values())
  .filter((p) => p.memory && p.prompt)   // <- one-sided pairs discarded
```

Items that referenced a missing counterpart were neither in `pairs` nor added to `standalone`.

## Fix

One-sided pairs now fall back to `standalone`, so orphaned linked memories/prompts still render, unlinked, in the timeline. Complete pairs still render together.

## Tests

- `memory timeline listing > keeps memories whose linked prompt is missing from the timeline` — reproduces the drop (fails before the fix, passes after)
- `memory timeline listing > renders linked memory-prompt pairs together in the timeline` — guards the pairing path

## Verification

- `tsc --noEmit` clean
- Full suite: 407 pass / 4 fail, identical failure set on clean `main` (pre-existing flakes: onnxruntime-resolve shim, plugin-loader contract timeout, turso-shard recreate timeout, config parallel). No new failures.
- Prettier clean

### DIFF
--- a/src/services/api-handlers.ts
+++ b/src/services/api-handlers.ts
@@ -248,9 +248,16 @@ export async function handleListMemories(
     }
 
     const sortedTimeline: any[] = [];
-    const pairs = Array.from(linkedPairs.values())
+    const pairValues = Array.from(linkedPairs.values());
+    const pairs = pairValues
       .filter((p) => p.memory && p.prompt)
       .sort((a, b) => b.memory.createdAt - a.memory.createdAt);
+    // A memory or prompt whose counterpart is missing (linked prompt deleted,
+    // or prompt capture off) must still show up in the timeline, unlinked.
+    for (const pair of pairValues) {
+      if (pair.memory && !pair.prompt) standalone.push(pair.memory);
+      else if (pair.prompt && !pair.memory) standalone.push(pair.prompt);
+    }
     for (const pair of pairs) {
       sortedTimeline.push(pair.memory);
       sortedTimeline.push(pair.prompt);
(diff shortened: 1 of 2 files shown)


lindixu6-hash (approved):
Reviewed the timeline pairing path and the new regression coverage. The change preserves complete memory/prompt ordering and moves only one-sided pairs into the existing timestamp-sorted standalone list.

Local verification on `c7b705fe6073517aa2611ea118f57d8dc1d1a232`: targeted orphan/pair tests `2 pass, 0 fail`; `bun run typecheck` passed; `bun run format:check` passed. The sandbox reported cleanup restrictions for temporary directories after the commands completed, so those restrictions did not affect the assertions or static checks.

lindixu6-hash:
Post-merge validation for exact merge commit `93cb1cf81049cbed785837507bc4b1a57445e5a9`: [Platform Package Smoke run 32276470861](https://github.com/lindixu6-hash/opencode-mem/actions/runs/32276470861) passed all six jobs (Ubuntu, Windows, macOS 15/26 on Apple Silicon, and macOS 15/26 Intel). Each job completed dependency installation, typecheck, production build, the full test suite, package creation/install, and native/runtime smoke checks.

PR #266, opened 2026-08-22
Include user-scope shards in all-projects tool queries

JojiiOfficial:
Tool search/list resolved "all-projects" to project shards only, so _user-scope_ memories were invisible to the memory tool. The API listing endpoint got the same fix earlier but the client path was missed.
`resolveMemoryScope` now returns both scopes and the client walks all of them.

### DIFF
--- a/src/services/client.ts
+++ b/src/services/client.ts
@@ -4,7 +4,11 @@ import { tursoVectorSearch } from "./turso/vector-search.js";
 import { tursoConnectionManager } from "./turso/connection-manager.js";
 import { ensureTursoReady } from "./turso/ready.js";
 import { formatTagsForEmbedding } from "./turso/vector-utils.js";
-import { extractScopeFromContainerTag, resolveMemoryScope } from "./memory-scope.js";
+import {
+  extractScopeFromContainerTag,
+  resolveMemoryScope,
+  type MemoryScopeRef,
+} from "./memory-scope.js";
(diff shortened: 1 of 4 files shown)


lindixu6-hash (approved):
Reviewed locally on the PR head (2841b5f).

**Correctness**
- The bug is real: `resolveMemoryScope("all-projects")` previously returned only `{scope:"project",hash:""}`, so the tool path (client.ts) walked project shards only while user-scope memories were silently excluded. The API listing endpoint (api-handlers.ts) already queried both scopes, confirming the intended behavior.
- `containerTag === ""` correctly means no container filter in both searchInShard (vector-search.ts:148-173) and listMemories (vector-search.ts:380), so user shards return user memories and project shards return project memories with no cross-contamination.
- No other callers of resolveMemoryScope exist; searchMemoriesBySessionID correctly uses extractScopeFromContainerTag directly (session-scoped, not affected).

**Testing**
- New integration tests verify getAllShards is called with both {scope:"user"} and {scope:"project"} for both listMemories and searchMemories under all-projects, and only project for project scope.
- Unit test updated for the new array return shape.
- 12/12 targeted tests pass; tsc --noEmit is clean.

**Nit (non-blocking)**
- resolveScopeValue is now a trivial pass-through to resolveMemoryScope. Could be inlined in a future cleanup but not worth holding this PR.

Good fix. Merging.

PR #267, opened 2026-08-24
fix: preserve provider errors in user profile learning

### LINKED ISSUE #265: fix: preserve provider errors in user profile learning
## Problem

`src/services/user-memory-learning.ts` appears to have the same error-masking pattern fixed for auto-capture in #258: when the OpenCode structured-output path fails and no manual external-API fallback is configured, the provider root cause can be replaced by a generic configuration error.

The adjacent auto-capture log at `src/services/auto-capture.ts:578` can also say that it is falling back to an external API when no fallback is configured.

## Expected behavior

- Preserve and surface the original provider error when profile learning has no configured manual fallback.
- Keep the configured manual fallback path unchanged.
- Ensure fallback logging accurately reflects whether a fallback will run.
- Add focused regression coverage for the no-fallback and configured-fallback paths.

## Context

Identified during the independent review of #258. Kept separate so that the already-reviewed auto-capture fix remains narrowly scoped.

lindixu6-hash:
## Summary

When the opencode provider path in `analyzeUserProfile()` throws and no manual external API fallback (`memoryModel`/`memoryApiUrl`) is configured, the original provider error was silently replaced by a generic `"External API not configured for user memory learning"` message. This made debugging provider failures (e.g. unsupported `tool_choice` in thinking mode) impossible from logs.

## Changes

- Store the opencode provider error in a local variable (same pattern as #258 for auto-capture).
- When no manual fallback is configured, rethrow the original error instead of the generic message.
- When a manual fallback IS configured, behavior is unchanged: the error is logged and the fallback runs.
- Added regression test `user-profile-learning-error.test.ts` that mocks a throwing opencode provider with no fallback and verifies the surfaced error contains the real root cause.

## Testing

- `bun test tests/user-profile-learning-error.test.ts` — 1/1 pass
- `tsc --noEmit` clean

Closes #265

### DIFF
--- a/src/services/user-memory-learning.ts
+++ b/src/services/user-memory-learning.ts
@@ -632,6 +632,7 @@ async function analyzeUserProfile(
   existingProfile: UserProfile | null
 ): Promise<AnalysisResult | null> {
   log("user-profile-learning: analyze called", { hasProfile: !!existingProfile });
+  let opencodeProviderError: unknown;
   if (CONFIG.opencodeProvider && CONFIG.opencodeModel) {
     log("user-profile-learning: trying opencode provider");
     try {
@@ -694,6 +695,7 @@ Use the update_user_profile tool to save the ${existingProfile ? "updated" : "ne
       }
       return { raw: rawData, merged: null };
     } catch (e) {
+      opencodeProviderError = e;
       log("user-profile-learning: opencode provider failed, falling back to external API", {
         error: String(e),
       });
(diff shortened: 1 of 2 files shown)


lindixu6-hash:
**Reviewer quick guide:** This is a 3-line source change mirroring the exact pattern already merged in #258 (auto-capture error preservation, reviewed by @NaNomicon).

**Source diff** (`src/services/user-memory-learning.ts`):
1. Declare `opencodeProviderError` before the try block (line 635)
2. Capture the caught error (line 698)
3. Rethrow it when no manual fallback exists (lines 706-708)

**Test:** `tests/user-profile-learning-error.test.ts` uses the same subprocess mock pattern as #258's test — mocks a throwing `generateStructuredOutput`, runs `performUserProfileLearning`, and asserts the surfaced error contains the real provider message ("Thinking mode does not support") rather than the generic "External API not configured".

**Risk:** Minimal — only changes the error message surfaced to users when their primary provider is broken and no fallback exists. The happy path and fallback path are untouched.

PR #269, opened 2026-08-24
test: add result-level ANN plan validation above k threshold

lindixu6-hash:
## Summary

Follow-up from #259 review feedback by @NaNomicon: the existing SQL-text assertion in `turso-vector-search.test.ts` only verifies the query contains `CROSS JOIN` and string ordering, which would pass even if the planner silently reordered and search degraded.

This adds a behavior-level test:
- Seeds **200 real memories** across two container tags (150 target + 50 other), well above the `k=128` ANN threshold for tagged searches.
- The target memory at index 0 has the query vector itself → must rank first with similarity > 0.9.
- Verifies **no other-container-tag memories leak** into results (indices 150–199 must not appear).
- Verifies result count respects the requested limit.

## Testing

- `bun test tests/turso-vector-search.test.ts` — 3/3 pass (existing 2 + new 1)
- `tsc --noEmit` clean

### DIFF
--- a/tests/turso-vector-search.test.ts
+++ b/tests/turso-vector-search.test.ts
@@ -100,4 +100,71 @@ describe("turso vector search", () => {
     }
     expect(observedSql[1]).toContain("m.container_tag = ?");
   });
+
+  it("returns correct tagged memories from ANN above the k threshold (result-level plan check)", async () => {
+    baseDir = mkdtempSync(join(tmpdir(), "turso-vector-ann-"));
+
+    const { CONFIG } = await import("../src/config.js");
+    CONFIG.storagePath = baseDir;
+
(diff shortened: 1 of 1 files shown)


lindixu6-hash:
**Reviewer quick guide:** This is a test-only PR (no source changes). It adds one integration test that seeds 200 real memories across two container tags (above the k=128 ANN threshold) and verifies:
1. The closest target-tag memory ranks first (similarity > 0.9)
2. No other-tag memories leak into results
3. Result count respects the limit

This addresses @NaNomicon's follow-up from #259 review: the existing SQL-text test only checks for `CROSS JOIN` in the query string, which would pass even if the planner reordered. This test validates actual ANN behavior at the result level.

PR #270, opened 2026-08-24
feat: show warning toast when opencode provider fails but fallback is configured

lindixu6-hash:
## Summary

When the opencode provider throws and a manual external API fallback is configured, the error was previously only logged to console. The user had no indication that their primary provider was broken and auto-capture was silently using the fallback.

This adds a non-blocking warning toast (5s, variant `warning`) informing the user that the opencode provider failed and the fallback is being used. The original provider error is included (truncated to 100 chars) for debugging.

## Changes

- `generateSummary` now accepts `ctx` to access `client.tui.showToast`.
- After the opencode provider catch block, if `opencodeProviderError` exists AND a manual fallback (`memoryModel` + `memoryApiUrl`) is configured AND `showErrorToasts` is enabled, a warning toast is shown.
- The error toast path (no fallback configured) from #258 is unchanged.
- When no provider error occurs, no toast is shown (existing behavior).

## Testing

- `tsc --noEmit` clean
- Existing auto-capture tests pass (2/2)
- Memory scope tests pass (4/4)

Follow-up from #259 review by @NaNomicon.

### DIFF
--- a/src/services/auto-capture.ts
+++ b/src/services/auto-capture.ts
@@ -107,7 +107,7 @@ async function capturePrompt(
 
         let summaryResult: { summary: string; type: string; tags: string[] } | null;
         try {
-          summaryResult = await generateSummary(context, sessionID, prompt.content, prompt);
+          summaryResult = await generateSummary(ctx, context, sessionID, prompt.content, prompt);
         } catch (error) {
           const message = error instanceof Error ? error.message : String(error);
           throw new Error(`Summary generation failed: ${message}`);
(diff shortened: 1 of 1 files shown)


lindixu6-hash:
**Reviewer quick guide:** This PR adds a non-blocking warning toast when the opencode provider fails but a manual fallback is configured. Previously the error was only logged, leaving users unaware their primary provider was broken.

**Source diff** (`src/services/auto-capture.ts`, ~20 lines):
- `generateSummary` now accepts `ctx` to access `showToast`
- After catching an opencode provider error, if a manual fallback exists AND `showErrorToasts` is enabled, show a 5s warning toast with the truncated error
- The error-rethrow path from #258 (no fallback configured) is unchanged
- The success path (no error) is unchanged

This addresses @NaNomicon's second follow-up from #259 review.

PR #276, opened 2026-08-25
fix: strip matching Tags footer only during compaction restore

phil-lipp:
Auto-capture embeds a "Tags: ..." footer inside the memory body (added in #131); compaction restore appends its own canonical Tags line, duplicating the tags in the restored context.

Strip a trailing `Tags:` footer only when its normalized value matches the structured tags, so auto-capture duplicates disappear without dropping user-authored `Tags:` content on manual/API/imported memories.

Replaces #249 (same change; that PR retained stale author metadata from earlier force-pushed commits).

Regression from #131.

## Test plan
- [x] Compaction tests (isolated HOME)
- [x] Mutation: greedy strip fails the unmatched-footer regression
- [x] Typecheck, production build, full suite after build

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1041,14 +1041,55 @@ function formatSearchResults(query: string, results: any, limit?: number): strin
   });
 }
 
+const EMBEDDED_TAGS_FOOTER_RE = /\n*Tags: ([^\n]*)\s*$/;
+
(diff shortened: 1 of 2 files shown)


phil-lipp:
Opened by mistake; the change stays on #249.

Issue #277, opened 2026-08-25
Vector index bloat: no compress_neighbors/max_neighbors tuning - 542 MB shard for ~656 memories (2048-dim embeddings)


shdowzh888:
### Environment

- opencode-mem: 2.22.2 (issue likely affects all versions up to 2.25.0)
- OS: Windows 11
- Embedding: remote OpenAI-compatible API (`doubao-embedding-vision`), **2048 dims**

### Symptom

With only ~656 memories in a project shard, the shard DB grew to **542 MB**. `dbstat` shows the two DiskANN shadow tables account for **98%** of the file:

| object | size |
|---|---|
| `memories_vec_idx_shadow` | 267.4 MB |
| `memories_tags_vec_idx_shadow` | 266.5 MB |
| `memories` (all actual text + metadata) | 11.0 MB |

That is ~850 KB of index overhead **per memory**, while the raw vectors are only 2048 × 4 B ≈ 8 KB each.

### Root cause

`initShardDb()` creates both vector indexes with `'metric=cosine'` only:

```sql
CREATE INDEX ... ON memories (libsql_vector_idx(vector, 'metric=cosine'))
```

libSQL defaults are aggressive at high dimensionality:

- `max_neighbors = 3·√D` → **~137 neighbors per node** at D=2048
- `compress_neighbors` unset → neighbor vectors stored as **uncompressed F32**

This matches the documented space formula `N · (Storage(T1) + M · Storage(T2))` in the [Turso docs](https://docs.turso.tech/features/ai-and-embeddings).

### Fix verified locally

Dropping and re-creating the same-named indexes with tuned parameters, then `VACUUM`, shrinks the whole shard from **542.5 MB → 73.3 MB (~7.4×)** with search quality intact:

```sql
DROP INDEX IF EXISTS memories_vec_idx;
DROP INDEX IF EXISTS memories_tags_vec_idx;
CREATE INDEX memories_vec_idx ON memories (libsql_vector_idx(vector,
  'metric=cosine', 'compress_neighbors=float8', 'max_neighbors=20'));
CREATE INDEX memories_tags_vec_idx ON memories (libsql_vector_idx(tags_vector,
  'metric=cosine', 'compress_neighbors=float8', 'max_neighbors=20'));
VACUUM;
```

[...]

PR #279, opened 2026-08-28
fix: reduce DiskANN index storage overhead

amandeavor:
## Summary

- store DiskANN neighbor vectors as `float8` for both content and tag indexes
- cap `max_neighbors` at 20 so high-dimensional embeddings do not inherit the much larger libSQL default
- assert the persisted SQL definitions and exercise direct `vector_top_k` lookups through both tuned indexes

## Scope

This deliberately changes newly created indexes only. Existing shards retain their current index definitions until a separate failure-safe rebuild/VACUUM migration is designed, so this patch does not risk leaving an existing shard without an index.

Addresses the new-shard portion of #277.

## Validation

- `bun test tests/turso-vector-search.test.ts -t "inserts and searches memories"` — 1 pass
- `bun run typecheck` — pass
- `bun run build` — pass
- full local Windows suite — 446 pass; 2 unrelated platform/timing failures: an `EBUSY` file rename in the legacy migrator test and the existing 30-second 200-vector ANN test timing out at 35.5 seconds. The latter was slower and eventually crashed Bun when repeated with the unmodified upstream index definition.

### DIFF
--- a/src/services/turso/shard-manager.ts
+++ b/src/services/turso/shard-manager.ts
@@ -328,13 +328,23 @@ export class TursoShardManager {
       {
         sql: `
           CREATE INDEX IF NOT EXISTS memories_vec_idx
-          ON memories (libsql_vector_idx(vector, 'metric=cosine'))
+          ON memories (libsql_vector_idx(
+            vector,
+            'metric=cosine',
+            'compress_neighbors=float8',
+            'max_neighbors=20'
+          ))
         `,
       },
       {
         sql: `
           CREATE INDEX IF NOT EXISTS memories_tags_vec_idx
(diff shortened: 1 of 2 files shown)


lindixu6-hash (approved):
Reviewed the new-shard index definition and direct-query coverage. The change is intentionally scoped to newly created content/tag indexes; existing-shard rebuild and VACUUM remain a separate migration concern. Local verification at `88221cd`: targeted vector tests 3/3, full suite 447/447 with 1 skip, typecheck, and production build all pass. The approved GitHub package-smoke matrix is also green on Ubuntu, Windows, macOS Intel, and Apple Silicon.

PR #289, opened 2026-08-30
fix: forward provider parameters during profile cleanup

### LINKED ISSUE #287: Forward memoryExtraParams in external AI cleanup requests
Split from #284 so provider request behavior can be fixed independently.

The external API path in `src/services/user-profile/ai-cleanup.ts` builds its request body directly and does not include configured `memoryExtraParams`. Provider-specific parameters such as `enable_thinking: false` therefore work in other model calls but are omitted during cleanup.

Acceptance criteria:
- forward configured extra parameters through the cleanup request path
- preserve required cleanup fields and prevent unsafe overrides
- add request-body coverage

See #284 for the SiliconFlow/Qwen reproduction.

amandeavor:
## Summary

Fixes #287.

The external profile-cleanup request now forwards configured memoryExtraParams through the same safe-parameter helper used by the other provider paths. Cleanup-critical fields are assigned afterward, so extra parameters cannot replace the model, messages, temperature, or required JSON response format.

## Validation

- bun test tests/ai-cleanup.test.ts: 8 passed
- bun run typecheck: passed
- production build: passed
- plugin bundle and loader contracts: 10 passed
- targeted Prettier check and git diff check: passed

The full Windows suite reached 437 passing tests. Two unrelated ONNX/vector tests timed out, and build-dependent tests initially failed before dist was generated; all 10 build-dependent tests passed after the production build.

### DIFF
--- a/src/services/ai/providers/base-provider.ts
+++ b/src/services/ai/providers/base-provider.ts
@@ -25,6 +25,7 @@ const PROTECTED_KEYS = new Set([
   "input",
   "instructions",
   "conversation",
+  "stream",
 ]);
 
 export function applySafeExtraParams(
--- a/src/services/user-profile/ai-cleanup.ts
+++ b/src/services/user-profile/ai-cleanup.ts
@@ -6,6 +6,7 @@ import {
   EXTERNAL_PROFILE_CLEANUP_TIMEOUT_MS,
   OPENCODE_PROFILE_CLEANUP_TIMEOUT_MS,
 } from "../request-timeouts.js";
+import { applySafeExtraParams } from "../ai/providers/base-provider.js";
 
 export interface AICleanupResult {
   cleaned: UserProfileData;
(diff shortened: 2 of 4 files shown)


lindixu6-hash (changes requested):
Blocking edge case: this newly forwards arbitrary `memoryExtraParams` into the cleanup Chat Completions request, but `applySafeExtraParams()` does not protect `stream`. With `memoryExtraParams: { stream: true }`, the endpoint returns SSE while `callViaExternalAPI()` unconditionally calls `response.json()`, so cleanup deterministically fails. Please either add `stream` to the protected keys (preferred, since the existing provider consumers are also non-streaming) or force `stream: false` after applying extras, and add a regression assertion for this case. I locally verified the current head otherwise: targeted cleanup tests pass, full suite 447/447 passes with 1 skip, typecheck passes, and production build passes.

amandeavor:
Added \stream\ to \PROTECTED_KEYS\ in \pplySafeExtraParams\ so non-streaming endpoints across all providers and AI profile cleanup are protected from accidental SSE streaming injection.

Also added regression coverage:
- In \	ests/ai-cleanup.test.ts\, verified external API requests ignore \stream: true\ in \memoryExtraParams\ and keep \stream\ undefined.
- In \	ests/ai-provider-config.test.ts\, added unit tests for \pplySafeExtraParams\ asserting protected keys (\stream\, \model\, \messages\, \	ools\, \	ool_choice\, \	emperature\, \input\, \instructions\, \conversation\) cannot override core request structure.

Verified locally:
- \un test tests/ai-cleanup.test.ts tests/ai-provider-config.test.ts tests/minimax-provider.test.ts tests/anthropic-provider.test.ts tests/openai-chat-completion-provider.test.ts\ (55/55 passed)
- \un run typecheck\ (\	sc --noEmit\ passed cleanly)
- Prettier formatting passed

amandeavor:
Fixed the streaming edge case in the latest commit. stream is now protected centrally in pplySafeExtraParams(), so it cannot put any of the existing non-streaming provider consumers into SSE mode. The cleanup regression fixture now passes stream: true and asserts that it is omitted from the final request body. Targeted cleanup tests pass (8/8), typecheck passes, and the production build passes. The full Windows suite reached 441 passing with 7 unrelated portability/worktree/SQLite timing failures.

PR #290, opened 2026-08-30
fix: show matched memory count during search

### LINKED ISSUE #288: Show matched memory count while Web UI search is active
Split from #284 so the Web UI behavior can be fixed independently.

The memory explorer count continues to display the store-wide total while a search is active. It should display the number of matched results so users can tell whether the current query found anything.

Acceptance criteria:
- show the filtered result count while search is active
- restore the full count when search is cleared
- add UI or state-level regression coverage

See #284 for the original reproduction.

amandeavor:
## Summary
- show the search result total while a query is active
- restore the store-wide total when the query is cleared
- cover matched, zero-result, and cleared-search states with regression tests

Closes #288

## Validation
- bun test tests/memory-count.test.ts (2 passed)
- bun run typecheck
- production web build
- Prettier check on changed files
- git diff --check

### DIFF
--- a/tests/memory-count.test.ts
+++ b/tests/memory-count.test.ts
@@ -1,0 +1,13 @@
+import { describe, expect, test } from "bun:test";
+import { getDisplayedMemoryCount } from "../web/src/lib/memory-count";
+
+describe("getDisplayedMemoryCount", () => {
+  test("shows matched results while search is active", () => {
+    expect(getDisplayedMemoryCount(true, 3, 120)).toBe(3);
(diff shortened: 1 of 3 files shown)


lindixu6-hash (approved):
Reviewed the search/count state flow and the extracted helper. The active-search path uses the API result total, clearing search restores the store-wide stats total, and zero-result search remains distinguishable from the unfiltered state. Local verification: targeted test 2/2, full suite 449/449 with 1 skip, typecheck, and production build all pass at `9fca346`.

PR #291, opened 2026-08-30
fix: preserve long-running cleanup requests

### LINKED ISSUE #285: AI cleanup aborts on premature request close event under Node.js 26
Split from #284 so this can be reproduced and fixed independently.

On Windows 11 with Node.js 26, AI cleanup can abort shortly after the request body is consumed. The request `close` event fires before the actual connection ends, and the current cleanup handler treats it as a client disconnect. The browser receives `ERR_EMPTY_RESPONSE` while server-side cleanup is still running.

Acceptance criteria:
- do not abort cleanup merely because the request body has closed
- still cancel work on a genuine client disconnect
- add a regression test for the request lifecycle

See #284 for the original environment and reproduction details.

### LINKED ISSUE #286: Align HTTP and AI cleanup timeout hierarchy
Split from #284 so the timeout policy can be handled independently.

The HTTP server timeout is 30 seconds while the external AI cleanup fetch allows 60 seconds. A valid slow cleanup can therefore be terminated by the outer server before the inner operation reaches its own deadline, producing an empty response.

Acceptance criteria:
- make the outer HTTP timeout longer than the cleanup operation deadline, or derive both from one setting
- document the timeout relationship
- cover the ordering with a focused test

See #284 for the observed timeout layers and reproduction.

lindixu6-hash:
## Summary

- stop treating a fully consumed Node request body `close` event as a client disconnect
- abort the fetch-style request only on a genuine request abort, premature response close, or socket failure/close
- derive the Node HTTP idle timeout from the longest profile-cleanup deadline with a 30-second transport grace period
- cover normal body completion, genuine disconnect deduplication, completed responses, and timeout ordering

Fixes #285
Fixes #286

## Validation

- `HOME=/tmp/opencode-mem-lifecycle-test bun test` — 453 passed, 1 skipped, 0 failed
- `bun run typecheck` — passed
- `bun run build` — passed
- targeted Prettier check — passed
- `git diff --check origin/main...HEAD` — passed
- Node.js v26.8.1 lifecycle probe reproduced the report: a fully consumed POST emits `req.close` with `req.complete=true` while `res.writableEnded=false`; the new guard ignores that event
- GitHub package-smoke matrix — Ubuntu, Windows, macOS Intel, and Apple Silicon all passed

## Boundary

The regression tests exercise the extracted lifecycle policy deterministically. The Node.js 26 probe verifies the runtime event ordering that triggered the original empty response.


amandeavor (approved):
LGTM! Tested locally on Windows (Node v24.16.0 / Bun 1.4.0):
- \un test tests/web-server-request-lifecycle.test.ts tests/ai-cleanup.test.ts\ (12/12 passed)
- \un run typecheck\ (\	sc --noEmit\ passed cleanly)

The extraction of \ttachNodeDisconnectHandlers\ cleanly handles the Node 26 \
eq.close\ event order after full body consumption, and coupling the transport timeout (\NODE_HTTP_IDLE_TIMEOUT_MS\) to the longest cleanup deadline plus grace period prevents premature aborts on large profiles.

Issue #173, opened 2026-07-17
OpenCode V2 plugin API migration: timeline and plan?


junyuyuan:
## Context

I'm a heavy user of opencode-mem (v2.19.4) on OpenCode V1 (1.18.3), relying on:
- Cross-session project memory (SQLite + vector search)
- Auto-capture and user profile learning
- memory tool integration in agent workflows
- Web UI for memory browsing

I noticed PR #101 already migrated the internal AI calls to OpenCode V2 SDK (session.prompt), which is great progress. However, OpenCode V2's [migration guide](https://v2.opencode.ai/migrate-v1) states:

> V1 plugins will not work in V2. The V2 plugin API is still being finalized during beta.

## Questions

1. **Is there a planned timeline for opencode-mem to fully support the OpenCode V2 plugin API?** Given that the internal SDK is already using V2 APIs, I'm hoping the plugin-level migration is on the near-term roadmap.

2. **What will the migration path look like for existing users?** Will existing memory data (SQLite databases) transfer seamlessly? Will the config format (opencode-mem.jsonc) remain compatible?

3. **Are there any V2-native features you're considering** (e.g., deeper integration with V2's session tree, agent-specific memory scopes)?

## My Setup

- OpenCode V1 1.18.3 on Windows
- opencode-mem v2.19.4 with custom DeepSeek V4 Flash endpoint for memory extraction
- ~50+ project memory records across multiple Roblox game projects
- Currently blocked from migrating to V2 until opencode-mem is ready

Thanks for the great work — cross-session memory is a critical part of my workflow.

EyJunge1:
Hey — the OpenCode V2 plugin API is still in beta, so endpoints and surfaces may still change. I don't want to build the production plugin path on an unstable API yet.

I've pushed a quick spike / test implementation to a separate branch: [`api-v2`](https://github.com/tickernelz/opencode-mem/tree/api-v2) (related research commit: [`3a80db1`](https://github.com/tickernelz/opencode-mem/commit/3a80db1b0a33fb2d485c49f7b391cbaa693b3c14)).

Once the V2 plugin API is stable, I'll update that branch and merge it into `main`. Until then we have to wait — V1 remains the supported path.

EyJunge1:
Quick correction: the spike branch is named [`opencode-api-v2`](https://github.com/tickernelz/opencode-mem/tree/opencode-api-v2) (renamed from `api-v2`).

danielgaribaldi:
[...]

PR #310, opened 2026-09-16
feat: add OpenCode v2 plugin entrypoint

peeyush-ciridae:
## Summary

- add a native OpenCode v2 entrypoint at `opencode-mem/v2`
- adapt memory tools, prompt/context injection, internal generation, and lifecycle events to the v2 APIs
- filter the global v2 event stream by plugin location
- clear per-session injected context before each prompt to avoid stale context reuse
- preserve the existing v1 entrypoint and document v2 configuration

## Validation

- `npm run typecheck`
- TypeScript build
- v1 and v2 loader contract tests: 8 tests passed
- package dry run confirmed `dist/v2/plugin.js` is published

Related to #173.


EyJunge1:
Thanks for putting this together.

I already have a dedicated V2 migration branch (`opencode-api-v2`, see #173) and I'll finish that path myself against the current v2 endpoints. I'll open a PR from there shortly. It's simpler to keep the migration in one place.

I'll close this one to avoid duplicate work. Appreciate the contribution.

